PR #6, opened 2019-12-05
Adding Subroutines and cycle support to Forge.

TravisOnGit:
Features:
1.	Subroutines
    - Subroutine support has been added. This native Forge Action walks a separate tree session and returns the last ActionResponse of that session.
    - ForgeTree
        - RootTreeNodeKey - property added. This specifies the TreeNodeKey that this ForgeTree should visit first, or "Root" by default.
        - Subroutine - TreeNodeType enum added. This node type must contain at least one SubroutineAction. Parallel SubroutineActions and regular Actions are supported.
    - SubroutineInput
        - TreeName - maps to a known ForgeTree in the App.
        - TreeInput - is a dynamic object that can be referenced from the schema.
    - InitializeSubroutineTree
        - This new Func is called in the SubroutineAction by Forge to get an initialized TreeWalkerSession from the App.
        - This gives fine-grain control to the App over the subroutine session that gets walked.
        - App is expected to initialize with a jsonSchema matching the TreeName, among other things.
        - App could choose to change any of the parameters, such as UserContext or ForgeActionsAssembly for different ForgeTrees if they desire.
2.	State Persistence and State Separation
    - Persisted state in Subroutine sessions are completely separate from the root session.
    - This means you cannot GetOutput from a node outside your session, for example.
    - State separation is accomplished by introducing RootSessionId. RootSessionId is kept the same across root and subroutine sessions.
    - IForgeDictionary.UpdateKeyPrefix method added that takes in both RootSessionId and SessionId to create a unique key per session. <RootSessionId>_<SessionId>_
    - Subroutine's SessionId is created and persisted by Forge in the SubroutineAction.
    - RootSessionId and TreeName are now passed along in ActionContext and ITreeWalkerCallbacks.
    - TreeName + TreeNodeKey can be used to uniquely identify a TreeNode at the global level. (Used in override debug APIs for example)
    - TreeName + TreeNodeKey + SessionId can be used to uniquely identify records at the session level. (Used in rate limiting for example)
3.	Cycles
    - Support for cycles (or revisiting a node) has been added. Code change for this is mostly in CommitCurrentTreeNode.
    - When revisiting a node, Forge will nullify the ActionResponses and Intermediates.
    - PreviousActionResponse persisted state has been added, and is set at the same time.
    - GetPreviousActionResponse() method added to ActionContext to allow actions access to that state.
4.	Misc
    - ForgeSchemaValidationTests added to validate the schemas used in UnitTests.
    - ForgeSchemaValidationRules.json updated with Subroutine support. Currently the SubroutineTypeNodeDefinition incorrectly allows a Subroutine type node to not include a SubroutineAction. I will fix this later.
    - EvaluateDynamicProperty now supports IDictionary types! It now also supports object types.

### DIFF
--- a/Forge.TreeWalker.UnitTests/test/ExecutorUnitTests.cs
+++ b/Forge.TreeWalker.UnitTests/test/ExecutorUnitTests.cs
@@ -16,6 +16,7 @@ namespace Forge.TreeWalker.UnitTests
 
     using Forge.ExternalTypes;
     using Forge.TreeWalker;
+    using Newtonsoft.Json.Linq;
 
     [TestClass]
     public class ExecutorUnitTests
@@ -43,7 +44,7 @@ namespace Forge.TreeWalker.UnitTests
         public void TestExecutor_Success_bool()
         {
             this.UserContext.Foo = "Bar";
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             string expression = "UserContext.Foo == \"Bar\"";
 
             Assert.IsTrue(ex.Execute<bool>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate a true expression.");
@@ -54,7 +55,7 @@ namespace Forge.TreeWalker.UnitTests
         {
             // Note that long requires casting where int does not.
             this.UserContext.Foo = (long)1000;
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             string expression = "UserContext.Foo";
 
             Assert.AreEqual((long)1000, ex.Execute<long>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate the expression.");
@@ -64,7 +65,7 @@ namespace Forge.TreeWalker.UnitTests
         public void TestExecutor_Success_int()
         {
             this.UserContext.Foo = 1000;
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             string expression = "UserContext.Foo";
 
             Assert.AreEqual(1000, ex.Execute<int>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate the expression.");
@@ -74,7 +75,7 @@ namespace Forge.TreeWalker.UnitTests
         public void TestExecutor_Success_ExecuteTwice()
         {
             this.UserContext.Foo = "Bar";
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             string expression = "UserContext.Foo == \"Bar\" && UserContext.GetTopic(\"TopicName\").ResourceType == \"Node\"";
 
             // Test - confirm Execute can compile and execute the same code twice without crashing.
@@ -86,7 +87,7 @@ namespace Forge.TreeWalker.UnitTests
         public void TestExecutor_Fail_MissingDefinitions()
         {
             this.UserContext.Foo = "Bar";
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             string expression = "UserContext.Bar == \"Bar\"";
 
             try
@@ -103,7 +104,7 @@ namespace Forge.TreeWalker.UnitTests
         public void TestExecutor_Fail_BadExpression()
         {
             this.UserContext.Foo = "Bar";
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             string expression = "UserContext.Foo";
 
             Assert.ThrowsException<InvalidCastException>(() => 
@@ -118,7 +119,7 @@ namespace Forge.TreeWalker.UnitTests
             this.UserContext.Foo = ExternalTestType.TestEnum;
             List<Type> dependencies = new List<Type>();
             dependencies.Add(typeof(ExternalTestType));
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, dependencies);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, dependencies, null);
             string expression = "UserContext.Foo.ToString() == ExternalTestType.TestEnum.ToString()";
             Assert.IsTrue(ex.Execute<bool>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate a true expression.");
         }
@@ -127,7 +128,7 @@ namespace Forge.TreeWalker.UnitTests
         public void TestExecutor_Fail_CompileExpressionWithMissingDependencies()
         {
             this.UserContext.Foo = ExternalTestType.ExampleEnum;
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             string expression = "UserContext.Foo == ExternalTestType.ExampleEnum";
 
             try
@@ -148,7 +149,7 @@ namespace Forge.TreeWalker.UnitTests
             List<Type> dependencies = new List<Type>();
             dependencies.Add(typeof(ExternalTestType));
 
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, dependencies);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, dependencies, null);
             string expression = "UserContext.Foo == ExternalTestType.TestEnum && UserContext.Bar == DiffNamespaceType.TestOne";
 
             try
@@ -169,7 +170,7 @@ namespace Forge.TreeWalker.UnitTests
             List<Type> dependencies = new List<Type>();
             dependencies.Add(typeof(ExternalTestType));
 
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, dependencies);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, dependencies, null);
             string expression = "UserContext.Foo.ToString() == ExternalTestType.TestEnum.ToString() && UserContext.Bar.ToString() == TestType.Test.ToString()";
             Assert.IsTrue(ex.Execute<bool>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate a true expression.");
         }
@@ -185,7 +186,7 @@ namespace Forge.TreeWalker.UnitTests
             dependencies.Add(typeof(Type));
 
             // Default dependencies are expected to be tossed away internally in ExpressionExecutor. 
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             string expression = "UserContext.Foo == \"Foo\"";
             Assert.IsTrue(ex.Execute<bool>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate a true expression.");
         }
@@ -193,7 +194,7 @@ namespace Forge.TreeWalker.UnitTests
         [TestMethod]
         public void TestExecutor_Success_ChangingFunctionDefinition()
         {
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
 
             // Rewritting GetCount to return 2
             string expression="UserContext.GetCount = new Func<int>(() => 2)";
@@ -205,7 +206,7 @@ namespace Forge.TreeWalker.UnitTests
         [TestMethod]
         public void TestExecutor_Fail_ChangingFunctionReturnType()
         {
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
 
             // Changing return type of GetCount
             string expression = "UserContext.GetCount = new Func<string>(() => \"Test\")";
@@ -219,7 +220,7 @@ namespace Forge.TreeWalker.UnitTests
         [TestMethod]
         public void TestExecutor_Success_ChangingFunctionReturnType()
         {
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
 
             // Changing return type of GetCount
             string expression = "UserContext.GetCount = new Func<string>(() => \"Test\")";
@@ -232,7 +233,7 @@ namespace Forge.TreeWalker.UnitTests
         [TestMethod]
         public void TestExecutor_Fail_ExecutingMultipleStatements()
         {
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
 
             // Changing return type of GetCount
             string expression = "int x = UserContext.GetCount() + 5; UserContext.GetCount = new Func<int>(() => x); return UserContext.GetCount()";
@@ -254,7 +255,7 @@ namespace Forge.TreeWalker.UnitTests
             string expression = "(Func<bool>)(() => {return UserContext.Foo == \"Bar\";})";
 
             // Casting the expression to Func<bool> since the executor will return a delegate of type Func<bool>
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             dynamic expressionResult = ex.Execute<Delegate>(expression).GetAwaiter().GetResult();
 
             if (expressionResult.GetType() == typeof(Func<bool>))
@@ -277,7 +278,7 @@ namespace Forge.TreeWalker.UnitTests
             string expression = "(Func<Task<bool>>)(() => {return Task.FromResult(UserContext.Foo == \"Bar\");})";
 
             // Casting the expression to Func<bool> since the executor will return a delegate of type Func<bool>
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             dynamic expressionResult = ex.Execute<Delegate>(expression).GetAwaiter().GetResult();
             
             if (expressionResult.GetType() == typeof(Func<Task<bool>>))
@@ -300,12 +301,45 @@ namespace Forge.TreeWalker.UnitTests
             string expression = "UserContext.Foo == \"Bar\" && UserContext.GetTopic(\"TopicName\").ResourceType == \"Node\"";
 
             // Test - confirm ExpressionExecutor script cache does not contain script before executing.
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             Assert.IsFalse(ex.ScriptCacheContainsKey(expression));
 
             // Test - confirm ExpressionExecutor script cache does contain script after executing.
             Assert.IsTrue(ex.Execute<bool>(expression).GetAwaiter().GetResult());
             Assert.IsTrue(ex.ScriptCacheContainsKey(expression));
         }
+
+        [TestMethod]
+        public void TestExecutor_TreeInput_Success()
+        {
+            object treeInput = "TestValue";
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, treeInput);
+            string expression = "TreeInput == \"TestValue\"";
+
+            Assert.IsTrue(ex.Execute<bool>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate a true expression using TreeInput.");
+        }
+
+        [TestMethod]
+        public void TestExecutor_TreeInputJObject_Success()
+        {
+            JObject treeInput = new JObject();
+            treeInput["StringProperty"] = "TestValue";
+            treeInput["IntProperty"] = 10;
+
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, treeInput);
+            string expression = "TreeInput.StringProperty == \"TestValue\" && TreeInput.IntProperty == 10";
+
+            Assert.IsTrue(ex.Execute<bool>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate a true expression using TreeInput as JObject.");
+        }
+
+        [TestMethod]
+        public void TestExecutor_TreeInputNull_Success()
+        {
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
+            string expression = "TreeInput == null";
+
+            Assert.IsTrue(ex.Execute<bool>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate a true expression using TreeInput as null.");
+        }
+
     }
 }
--- a/Forge.TreeWalker.UnitTests/test/ExecutorUnitTests.cs
+++ b/Forge.TreeWalker.UnitTests/test/ExecutorUnitTests.cs
@@ -44,7 +44,7 @@ namespace Forge.TreeWalker.UnitTests
         public void TestExecutor_Success_bool()
         {
             this.UserContext.Foo = "Bar";
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
+            ExpressionExecutor ex = new ExpressionExecutor(session: null, userContext: this.UserContext);
             string expression = "UserContext.Foo == \"Bar\"";
 
             Assert.IsTrue(ex.Execute<bool>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate a true expression.");
@@ -55,7 +55,7 @@ namespace Forge.TreeWalker.UnitTests
         {
             // Note that long requires casting where int does not.
             this.UserContext.Foo = (long)1000;
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
+            ExpressionExecutor ex = new ExpressionExecutor(session: null, userContext: this.UserContext);
             string expression = "UserContext.Foo";
 
             Assert.AreEqual((long)1000, ex.Execute<long>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate the expression.");
@@ -65,7 +65,7 @@ namespace Forge.TreeWalker.UnitTests
         public void TestExecutor_Success_int()
         {
             this.UserContext.Foo = 1000;
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
+            ExpressionExecutor ex = new ExpressionExecutor(session: null, userContext: this.UserContext);
             string expression = "UserContext.Foo";
 
             Assert.AreEqual(1000, ex.Execute<int>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate the expression.");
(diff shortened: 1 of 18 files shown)


Forge.TreeWalker/contracts/ForgeSchemaValidationRules.json
```diff
@@ -1,193 +1,281 @@
...
+                        {
+                            "not": {
+                                "patternProperties": {
+                                    ".*?": {
+                                        "$ref": "#/definitions/SubroutineActionDefinition"
+                                    }
+                                }
+                            }
+                        },
+                        {
```
abhayketkar: test comment

Forge.TreeWalker/contracts/ForgeSchemaValidationRules.json
```diff
@@ -1,193 +1,281 @@
...
-          },
-          "minItems": 1
+    "type": "object",
+    "properties": {
+        "Tree": {
+            "$ref": "#/definitions/TreeDefinition"
         },
-        "Properties": {
-          "type": "object"
+        "RootTreeNodeKey": {
```
sharanyas007: Test

Forge.TreeWalker/contracts/ForgeSchemaValidationRules.json
```diff
@@ -1,193 +1,281 @@
...
-          "enum": [ "Action" ]
-        },
-        "Timeout": {
-          "type": [ "number", "string" ]
-        },
-        "Actions": {
-          "patternProperties": {
-            ".*?": {
-              "$ref": "#/definitions/ActionDefinition"
+    "additionalProperties": false,
```
jamwAzureCompute: Travis is the best

Forge.TreeWalker/src/SubroutineAction.cs
```diff
@@ -0,0 +1,113 @@
...
+            // Initialize TreeWalkerSession for this subroutine.
+            TreeWalkerSession subroutineSession = this.parameters.InitializeSubroutineTree(input, intermediates.SessionId, this.parameters);
+
+            // Update KeyPrefix of ForgeState for state persistence separation.
+            subroutineSession.Parameters.ForgeState.UpdateKeyPrefix(subroutineSession.Parameters.RootSessionId, subroutineSession.Parameters.SessionId);
+
+            // Walk tree starting at CurrentTreeNode if persisted, otherwise RootTreeNodeKey if given, otherwise "Root" as default.
+            string currentTreeNode = await subroutineSession.GetCurrentTreeNode() ?? subroutineSession.Schema.RootTreeNodeKey;
+
+            // WalkTree may throw exceptions. We let this happen to allow for possible retry handling.
```
accioAzure: "possible retry handling" will the code of retry reside inside the forge tree walker or on the anvil side?
also what are cases where it will not retry?

TravisOnGit: Here, the retry handling refers to the native RetryPolicy in Forge for all Actions. You can read more about it here: https://github.com/microsoft/Forge/wiki/ForgeAction-Authoring---How-To-Guide#Forge-Behavior-and-how-it-changes-based-on-RunAction-response-and-ForgeSchema-configuration

Forge.TreeWalker/src/ExpressionExecutor.cs
```diff
@@ -49,13 +49,15 @@ public class ExpressionExecutor
         /// <param name="userContext">The dynamic user context.</param>
         /// <param name="dependencies">Type dependencies required to compile the schema. Can be null if no external dependencies required.</param>
         /// <param name="scriptCache">Script cache used to cache and re-use compiled Roslyn scripts.</param>
-        public ExpressionExecutor(ITreeSession session, object userContext, List<Type> dependencies, ConcurrentDictionary<string, Script<object>> scriptCache)
+        /// <param name="treeInput">The dynamic TreeInput object for this tree walking session.</param>
```
sharanyas007: Test comment 2

Forge.TreeWalker/contracts/ForgeTree.cs
```diff
@@ -25,6 +24,12 @@ public class ForgeTree
         /// </summary>
         [DataMember]
         public Dictionary<string, TreeNode> Tree { get; set; }
+
+        /// <summary>
+        /// The root TreeNodeKey that should be visited first when walking the tree.
+        /// </summary>
+        [DataMember]
```
abhayketkar: Test Comment

Forge.TreeWalker/src/ForgeDictionary.cs
```diff
@@ -19,11 +19,6 @@ namespace Forge.TreeWalker
     /// </summary>
     public class ForgeDictionary : IForgeDictionary
     {
-        /// <summary>
-        /// The unique identifier for this session.
-        /// </summary>
-        public Guid SessionId { get; set; }
```
abhayketkar: Is this SessionId not useful for the application? I see that Forge only needs the keyPrefix but the application might find it useful for logging/querying session information etc.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -56,6 +54,18 @@ public class TreeWalkerSession : ITreeSession
         /// </summary>
         public static string IntermediatesSuffix = "_Int";
 
+        /// <summary>
+        /// The TreeInput suffix appended to the end of the key in forgeState that maps to this tree walking session's TreeInput object.
+        /// </summary>
+        public static string TreeInputSuffix = "TI";
```
zhengmu: "TI" => "_TI" to follow the pattern used on other Suffix?

Forge.TreeWalker/src/ITreeWalkerCallbacks.cs
```diff
@@ -28,7 +26,9 @@ public interface ITreeWalkerCallbacks
         /// <param name="properties">The additional properties for this node.</param>
         /// <param name="userContext">The dynamic user-defined context object.</param>
         /// <param name="token">The cancellation token.</param>
-        Task BeforeVisitNode(Guid sessionId, string treeNodeKey, dynamic properties, object userContext, CancellationToken token);
+        /// <param name="treeName">The name of the ForgeTree in the JsonSchema.</param>
+        /// <param name="rootSessionId">The unique identifier for the root/parent tree walking session.</param>
+        Task BeforeVisitNode(Guid sessionId, string treeNodeKey, dynamic properties, object userContext, CancellationToken token, string treeName, Guid rootSessionId);
```
abhayketkar: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1068?view=vs-2019
CancellationToken should be the last parameter. This comment applies to all occurences of this.

Can we enable FxCop for Forge Projects?

Forge.TreeWalker/src/SubroutineAction.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// The tree walker parameters of the parent tree walker session.
+        /// Used to call InitializeSubroutineTree to get the initialized TreeWalkerSession for this Subroutine.
+        /// </summary>
+        private TreeWalkerParameters parameters { get; set; }
+
+        /// <summary>
+        /// Instantiates a SubroutineAction with the required parameters.
+        /// </summary>
+        /// <param name="parameters">The tree walker parameters of the parent tree walker session.</param>
+        public SubroutineAction(TreeWalkerParameters parameters)
```
abhayketkar: null check on parameters

Forge.TreeWalker/src/ForgeDictionary.cs
shweta241684: oh is it? I was under the impression that during rehydrate scenario, this helps in identifying the subroutine if they are running in paralel?

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -421,5 +439,237 @@ public static class ForgeSchemaHelper
                 }
             }
         ";
+
+        public const string SubroutineAction_GetLastActionResponse = @"
```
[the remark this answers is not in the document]
TravisOnGit: SubroutineAction_GetLastActionResponse is getting called in UTs. You can use "Peek References" if you are using VSCode extension to see that. You are also right that I am testing different schema configurations in the ForgeSchemaValidationTests as well.

Forge.TreeWalker.UnitTests/test/ForgeSchemaValidationTests.cs
```diff
@@ -0,0 +1,184 @@
...
+            this.jsonSchemaRules = File.ReadAllText(Path.Combine(Environment.CurrentDirectory, "contracts\\ForgeSchemaValidationRules.json"));
+            this.rules = JSchema.Parse(this.jsonSchemaRules);
+
+            // Update the blacklist of forgeTrees that are expected to fail validation for testing purposes.
+            // Key is the const string variable name representing a schema file in ForgeSchemaHelper.
+            // Value is the list of TreeNames in the schema if the schema has TreeNames. Or "NA" if the schema contains just a ForgeTree without TreeName.
+            this.jsonSchemaFailureBlacklist = new Dictionary<string, List<string>>()
+            {
+                {
+                    "SubroutineAction_FailsOnActionTreeNodeType",
```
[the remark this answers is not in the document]
TravisOnGit: Like it, done.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -13,17 +13,18 @@ public static class ForgeSchemaHelper
     {
         public const string TardigradeScenario = @"
             {
-                ""Tree"":
-                {
+                ""Tree"": {
```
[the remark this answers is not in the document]
TravisOnGit: Good point, I've made everything consistent now.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -421,5 +439,237 @@ public static class ForgeSchemaHelper
...
+                                    }
+                                }
+                            }
+                        }
+                    }
+                },
+                ""SubroutineTree"": {
+                    ""Tree"": {
+                        ""Root"": {
+                            ""Type"": ""Leaf""
```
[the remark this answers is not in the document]
TravisOnGit: Not sure exactly what you are asking, but the SubroutineAction is treated similarly to other Actions. The RunAction method is called and returns an ActionResponse. Forge persists this depending on what gets returned and the RetryPolicy in place. There is documentation for this in the wiki. If I didn't answer your question, please clarify, thanks.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -421,5 +439,237 @@ public static class ForgeSchemaHelper
...
+                    ""Tree"": {
+                        ""Root"": {
+                            ""Type"": ""Subroutine""
+                        }
+                    }
+                }
+            }
+        ";
+
+        public const string CycleSchema = @"
```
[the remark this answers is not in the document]
TravisOnGit: It is also used in UTs.

Forge.TreeWalker/src/TreeWalkerSession.cs
TravisOnGit: The tree-level persisted rows do not have "_" because the final key looks like this:
<RootSessionId>_<SessionId>_**TI**

Action-level persisted rows have the "_" because the final key looks like this:
<RootSessionId>_<SessionId>_<TreeActionKey>**_AR**

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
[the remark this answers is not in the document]
zhengmu: Let me make my question clear.

For regular Action node, if the Action is "Leaf" type, then the tree walk ends here. no question about it.

For the subroutine Tree, if the Node is "Leaf" type, will 1) the subrouting tree walk end and return to the parent tree, or 2) the whole tree walk ends here,?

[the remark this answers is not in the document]
TravisOnGit: 1. The Subroutine session's WalkTree will complete from within the SubroutineAction in the parent session. The parent session will continue.

Forge.TreeWalker.UnitTests/test/ExecutorUnitTests.cs
```diff
@@ -185,15 +186,15 @@ public void TestExecutor_Success_CompileExpressionWithForgeDefaultDependenciesBe
...
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
             string expression = "UserContext.Foo == \"Foo\"";
             Assert.IsTrue(ex.Execute<bool>(expression).GetAwaiter().GetResult(), "Expected ExpressionExecutor to successfully evaluate a true expression.");
         }
 
         [TestMethod]
         public void TestExecutor_Success_ChangingFunctionDefinition()
         {
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
```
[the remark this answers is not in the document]
TravisOnGit: Done. Added new overload to reduce parameters needed for most UTs.

Forge.TreeWalker.UnitTests/test/ExecutorUnitTests.cs
```diff
@@ -185,15 +186,15 @@ public void TestExecutor_Success_CompileExpressionWithForgeDefaultDependenciesBe
             dependencies.Add(typeof(Type));
 
             // Default dependencies are expected to be tossed away internally in ExpressionExecutor. 
-            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null);
+            ExpressionExecutor ex = new ExpressionExecutor(null, this.UserContext, null, null);
```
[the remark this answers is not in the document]
TravisOnGit: I've updated the overloads. I was initially avoiding adding scriptCache to the parameters because even if you pass in null, you need to reference all the Roslyn packages. But it made the overloads look weird.

I've bit the bullet on that and have included the Roslyn packages in UnitTests csproj. I've amended the overloads to look better.

Forge.TreeWalker/src/ITreeWalkerCallbacks.cs
TravisOnGit: Moved token to the end. I was initially thinking to not break backwards compatibility by changing the parameters order. But I'm breaking backwards compatibility anyways since these are new required parameters, so no harm in moving the token param to the end. :)

Forge.TreeWalker/src/SubroutineAction.cs
TravisOnGit: Yes, added.

Forge.TreeWalker/src/ForgeDictionary.cs
TravisOnGit: The app can implement IForgeDictionary however they like, including adding additional properties for SessionId if they want access to that directly. I didn't see a need for this property with the addition of UpdateKeyPrefix and GetKeyPrefix. The app could also string parse the GetKeyPrefix result.

Forge.TreeWalker/contracts/ForgeSchemaValidationRules.json
TravisOnGit: Test reply from codeflow.

---
In reply to: [355064064](https://github.com/microsoft/Forge/pull/6#discussion_r355064064) [](ancestors = 355064064)

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -56,6 +54,18 @@ public class TreeWalkerSession : ITreeSession
         /// </summary>
         public static string IntermediatesSuffix = "_Int";
 
+        /// <summary>
+        /// The TreeInput suffix appended to the end of the key in forgeState that maps to this tree walking session's TreeInput object.
+        /// </summary>
```
zhengmu: nit:     better if you have the example as the existing comments like
 /// Key: <SessionId>_CTN
or 
  /// Key: <SessionId>_<TreeActionKey>_Int

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -56,6 +54,18 @@ public class TreeWalkerSession : ITreeSession
...
+        public static string TreeInputSuffix = "TI";
+
+        /// <summary>
+        /// The PreviousActionResponse suffix appended to the end of the key in forgeState that maps to a previously persisted ActionResponse.
+        /// When a TreeNodeKey in a tree walking session was previously successfully visited, the ActionResponses get wiped and persisted to PreviousActionResponse.
+        /// GetPreviousActionResponse method is available in ActionContext.
+        /// </summary>
```
zhengmu: same nit comment as above

Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -81,6 +81,39 @@ public class TreeWalkerParameters
...
+        /// <summary>
+        /// The unique identifier of the root/parent session that gets passed on to Subroutine sessions.
+        /// RootSessionId will get set to SessionId if not initialized.
+        /// </summary>
+        public Guid RootSessionId { get; set; }
+
+        /// <summary>
+        /// The name of the ForgeTree in the JsonSchema.
+        /// For Subroutines, this is evaluated from the SubroutineInput on the schema.
+        /// "RootTree" will be used as TreeName if not specified.
```
zhengmu: >RootTree [](start = 13, length = 8)

suggest making "RootTree" as a public const in this class.

Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -81,6 +81,39 @@ public class TreeWalkerParameters
...
+
+        /// <summary>
+        /// The unique identifier of the root/parent session that gets passed on to Subroutine sessions.
+        /// RootSessionId will get set to SessionId if not initialized.
+        /// </summary>
+        public Guid RootSessionId { get; set; }
+
+        /// <summary>
+        /// The name of the ForgeTree in the JsonSchema.
+        /// For Subroutines, this is evaluated from the SubroutineInput on the schema.
```
zhengmu: >SubroutineInput on the schema [](start = 56, length = 29)

we probably need a Unittest to make sure the Subroutine's input cannot have "RootTree" as the tree name

Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -81,6 +81,39 @@ public class TreeWalkerParameters
...
+        public Func<SubroutineInput, Guid, TreeWalkerParameters, TreeWalkerSession> InitializeSubroutineTree { get; set; }
+
+        /// <summary>
+        /// The unique identifier of the root/parent session that gets passed on to Subroutine sessions.
+        /// RootSessionId will get set to SessionId if not initialized.
+        /// </summary>
+        public Guid RootSessionId { get; set; }
+
+        /// <summary>
+        /// The name of the ForgeTree in the JsonSchema.
```
zhengmu: >the [](start = 24, length = 3)

if you have a new iteration, you can change "the" => "this" to make it more clear.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -459,22 +498,73 @@ internal async Task PerformLeafTypeBehavior(TreeNode treeNode)
...
+        /// <param name="treeNode">The TreeNode containing actions to execute.</param>
+        /// <param name="treeNodeKey">The TreeNode's key where the actions are taking place.</param>
+        /// <exception cref="ArgumentException">If the pre-checks fail.</exception>
+        internal async Task PerformSubroutineTypeBehavior(TreeNode treeNode, string treeNodeKey)
+        {
+            // Perform pre-checks.
+            if (treeNode.Actions == null)
+            {
+                throw new ArgumentException("Subroutine TreeNodeTypes must contain at least one SubroutineAction. TreeNodeKey: " + treeNodeKey);
```
zhengmu: >must contain at least one SubroutineAction [](start = 70, length = 42)

suggest "must contain at least one SubroutineAction" => "doesn't contain any action". This is to make the fault message different than the one down below. Easy to tell which condition to trigger the fault.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -459,22 +498,73 @@ internal async Task PerformLeafTypeBehavior(TreeNode treeNode)
             }
         }
 
+        /// <summary>
+        /// Performs Subroutine TreeNodeType behavior.
+        /// </summary>
+        /// <param name="treeNode">The TreeNode containing actions to execute.</param>
+        /// <param name="treeNodeKey">The TreeNode's key where the actions are taking place.</param>
+        /// <exception cref="ArgumentException">If the pre-checks fail.</exception>
+        internal async Task PerformSubroutineTypeBehavior(TreeNode treeNode, string treeNodeKey)
```
zhengmu: >PerformSubroutineTypeBehavior [](start = 28, length = 29)

looks like this check function can be part of PerformActionTypeBehavior()

bool preCheck_ContainsAtLeastOneSubroutineAction = false;
bool preCheck_ContainsNoSubroutineActions = true;
...

if (treeNode.Type != TreeNodeType.Subroutine && !preCheck_ContainsNoSubroutineActions)
{throw...}

if (treeNode.Type == TreeNodeType.Subroutine && !ContainsAtLeastOneSubroutineAction )
{throw...}

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -884,47 +997,109 @@ public async Task<object> EvaluateDynamicProperty(dynamic schemaObj, Type knownT
...
+                foreach (KeyValuePair<string, TreeAction> kvp in treeNode.Actions)
+                {
+                    string treeActionKey = kvp.Key;
+                    TreeAction treeAction = kvp.Value;
+                    ActionResponse actionResponse = await this.GetOutputAsync(treeActionKey).ConfigureAwait(false);
+
+                    if (actionResponse == null)
+                    {
+                        // If this session has rehydrated and actionResponse is null, then this must be the first time visiting this TreeNode.
+                        itemsToPersist.Clear();
```
zhengmu: >itemsToPersist.Clear(); [](start = 24, length = 23)

if an actionNode has 2 actions; one has response saved, and the other one has null for response. Shall we clear for all?

And in this case, the value of itemsToPersist will be decided by which action to check further. If the has-responsed is checked first, then itemsToPersist will be empty. If the has-response is checked secondly, then itemsToPersist has the result of the has-responsed.

Forge.TreeWalker/src/TreeWalkerParameters.cs
TravisOnGit: Great suggestion, done.

---
In reply to: [356343301](https://github.com/microsoft/Forge/pull/6#discussion_r356343301) [](ancestors = 356343301)

Forge.TreeWalker/src/TreeWalkerParameters.cs
TravisOnGit: I think "the" makes more sense.

---
In reply to: [356344162](https://github.com/microsoft/Forge/pull/6#discussion_r356344162) [](ancestors = 356344162)

Forge.TreeWalker/src/TreeWalkerParameters.cs
TravisOnGit: A Subroutine can use the same ForgeTree as a parent session. If the application can prevent that with UTs if they wish.

---
In reply to: [356343697](https://github.com/microsoft/Forge/pull/6#discussion_r356343697) [](ancestors = 356343697)

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -884,47 +997,109 @@ public async Task<object> EvaluateDynamicProperty(dynamic schemaObj, Type knownT
...
+
+            itemsToPersist.Add(new KeyValuePair<string, object>(CurrentTreeNodeSuffix, treeNodeKey));
+
+            await this.Parameters.ForgeState.SetRange(itemsToPersist);
+            this.hasSessionRehydrated = true;
         }
 
         /// <summary>
-        /// Commits the last tree action to the forgeState.
+        /// Gets the previously committed TreeInput if it exists, otherwise persists the incoming TreeInput.
```
zhengmu: >Gets the previously committed TreeInput if it exists, otherwise persists the incoming TreeInput. [](start = 12, length = 96)

If the previously committed TreeInput is different than the incoming TreeInput, say some parameter value changes in the new schema. Shall we use the new one, which has the most recently requirement?

Forge.TreeWalker/src/TreeWalkerSession.cs
TravisOnGit: I want to give the TreeNodeTypes their own methods for extra clarity. If the logic was intertwined inside ActionTypeBehavior, it becomes more difficult to know what is special about Subroutine type.
This also allows to more easily add additional logic later that is specific for certain types.

---
In reply to: [356350218](https://github.com/microsoft/Forge/pull/6#discussion_r356350218) [](ancestors = 356350218)

Forge.TreeWalker/src/TreeWalkerSession.cs
TravisOnGit: This part is slightly tricky because it isn't intuitive why I can make this statement:
"If this session has rehydrated and actionResponse is null, then this must be the first time visiting this TreeNode"

But it turns out to be true! If I'm already rehydrated, then to get to this line of code, I have successfully visited a previous node and have childSelected to this node. So there are two possible states for this node:
1. All ActionResponses are persisted and we are revisiting this node.
2. All ActionResponses must be null, and we are visiting this node for the first time.

I try to leave enough comments around to describe this behavior, but let me know if I should add more.

---
In reply to: [356750599](https://github.com/microsoft/Forge/pull/6#discussion_r356750599) [](ancestors = 356750599)

Forge.TreeWalker/src/ForgeDictionary.cs
```diff
@@ -41,12 +36,12 @@ public class ForgeDictionary : IForgeDictionary
         /// ForgeDictionary Constructor.
         /// </summary>
         /// <param name="forgeStateTable">The forge state table dictionary object.</param>
+        /// <param name="rootSessionId">The unique identifier for the root/parent session.</param>
         /// <param name="sessionId">The unique identifier for this session.</param>
-        public ForgeDictionary(IDictionary<string, object> forgeStateTable, Guid sessionId)
+        public ForgeDictionary(IDictionary<string, object> forgeStateTable, Guid rootSessionId, Guid sessionId)
```
zhengmu: >Guid rootSessionId, Guid sessionId [](start = 76, length = 34)

do we need three tuple, rootSessionId (root tree), parentSessionId (for the caller of the sub-tree), and sessionId (for this sub-tree)?

scenario is TreeRoot calls TreeBranchA and TreeBranchB, while A and B can both call the subTreeC. With the two of rootSessionId and SessionId of subTreeC, we cannot tell either BranchA or BranchB calles the subTreeC, until we walk the whole sesssion from RootTree start.

Forge.TreeWalker/src/TreeWalkerParameters.cs
TravisOnGit: I added RootTree public static to TreeWalkerSession since that is where I am setting the value. It would also make sense to put it in this class instead. Let me know if you have stronger feelings to put it here instead.

---
In reply to: [356754558](https://github.com/microsoft/Forge/pull/6#discussion_r356754558) [](ancestors = 356754558,356343301)

Forge.TreeWalker/src/TreeWalkerSession.cs
TravisOnGit: No, we use the previous one intentionally. Otherwise there would be no need to persist the TreeInput.
Here is the reasoning for this decision. For example, you pass in a TreeInput, and you rely on the value of the TreeInput to walk to NodeA, and then failover.
Upon rehydration, let's say the TreeInput is calculated differently (perhaps indicating we should have walked to NodeB instead).
Because we are already at NodeA, updating the TreeInput could break our assumptions.
We want avoid that, so we persist the TreeInput and use that instead.

---
In reply to: [356762006](https://github.com/microsoft/Forge/pull/6#discussion_r356762006) [](ancestors = 356762006)

Forge.TreeWalker/src/ForgeDictionary.cs
TravisOnGit: The concept of state separation for different sessions works a bit differently than you describe. The subroutine sessions don't need to call back or know about their parent sessions.
The RootSessionId is passed along just so the application can keep track that all the Subroutine sessions belong to one uber-session.
Technically, the RootSessionId is not required to attain state separation.

The SubroutineAction creates a new Guid SessionId for its Subroutine TreeWalkerSession, and persists it in its Intermediates.
Let's say a Subroutine session contains another nested Subroutine session. Everything works the same, a new Guid SessionId is created.
The SessionId alone is enough to uniquely identify any session. You don't have to know anything about its parents.

---
In reply to: [356766078](https://github.com/microsoft/Forge/pull/6#discussion_r356766078) [](ancestors = 356766078)

Forge.TreeWalker/src/SubroutineAction.cs
```diff
@@ -0,0 +1,113 @@
...
+            SubroutineInput input = (SubroutineInput)actionContext.ActionInput;
+
+            // Rehydrate the subroutine's SessionId if previously persisted.
+            SubroutineIntermediates intermediates = await actionContext.GetIntermediates<SubroutineIntermediates>();
+            if (intermediates == null)
+            {
+                intermediates = new SubroutineIntermediates()
+                {
+                    SessionId = Guid.NewGuid()
+                };
```
zhengmu: >; [](start = 17, length = 1)

nit: move this ";" to the end of line above

TravisOnGit: I need the semi-colon here to end the line creating the SubroutineIntermediates object. I do not need semi-colon after initializing each parameter in this way (you use commas if there are multiple parameters).

---
In reply to: [356863479](https://github.com/microsoft/Forge/pull/6#discussion_r356863479) [](ancestors = 356863479)

PR #15, opened 2020-05-16
Adding Enum support

TravisOnGit:
While evaluating tree properties, Forge is now able to parse integer and string types into Enum members.

This required 3 small changes across EvaluateDynamicProperty and ExpressionExecutor.

Example of what Forge can now parse:
```json
"FooEnumArray": [
    "TestEnumTwo",
    "1",
    1,
    "C#|\"TestEnumTwo\"",
    "C#|\"1\"",
    "C#|1",
    "100",
    100,
    "C#|FooEnum.TestEnumTwo",
    "C#|(FooEnum)Enum.Parse(typeof(FooEnum), \"TestEnumOne\")"
]
```
```cs
    public enum FooEnum
    {
        TestEnumOne = 0,
        TestEnumTwo = 1
    }
```
Note that Forge will hit exceptions while parsing Enums if the value is: null, string.Empty, or a string that doesn't match a defined Enum member (e.g. "TestEnumOneHundred").

Note that integer (and integer as string) values that do not map to a defined Enum member are allowed. E.g. 100. This is because it is allowed in .Net and does not throw an exception. E.g. Console.WriteLine((FooEnum)100);.

### DIFF
--- a/Forge.TreeWalker/src/ExpressionExecutor.cs
+++ b/Forge.TreeWalker/src/ExpressionExecutor.cs
@@ -110,7 +110,10 @@ namespace Microsoft.Forge.TreeWalker
                              this.scriptOptions,
                              typeof(CodeGenInputParams)));
 
-            return (T)(await script.RunAsync(this.parameters).ConfigureAwait(false)).ReturnValue;
+            // Execute script and return the result.
+            // Parse Enum types explicitly since they cannot be casted directly.
+            var temp = (await script.RunAsync(this.parameters).ConfigureAwait(false)).ReturnValue;
+            return typeof(T).IsEnum ? (T)Enum.Parse(typeof(T), temp.ToString()) : (T)temp;
         }
 
         /// <summary>
--- a/Forge.TreeWalker/src/TreeWalkerSession.cs
+++ b/Forge.TreeWalker/src/TreeWalkerSession.cs
@@ -923,6 +923,12 @@ namespace Microsoft.Forge.TreeWalker
                         }
                     }
 
+                    if (knownType?.IsEnum ?? false)
+                    {
+                        // Case when schema property is an Enum type.
+                        return Enum.Parse(knownType, schemaObj);
+                    }
+
                     return schemaObj;
                 }
                 else if (schemaObj is JObject)
@@ -978,8 +984,20 @@ namespace Microsoft.Forge.TreeWalker
                 else
                 {
                     // Case when schema object is a value type.
+                    // Return the value as an Enum if knownType is an Enum.
                     // Return the value as the knownType if given.
-                    return knownType != null ? Convert.ChangeType(schemaObj, knownType) : schemaObj;
+                    if (knownType == null)
+                    {
+                        return schemaObj;
+                    }
+                    else if (knownType.IsEnum)
+                    {
+                        return Enum.Parse(knownType, schemaObj.ToString());
+                    }
+                    else
+                    {
+                        return Convert.ChangeType(schemaObj, knownType);
+                    }
                 }
             }
             catch (OperationCanceledException)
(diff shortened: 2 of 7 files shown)


Forge.TreeWalker/src/ExpressionExecutor.cs
```diff
@@ -110,7 +110,10 @@ public async Task<T> Execute<T>(string expression)
                              this.scriptOptions,
                              typeof(CodeGenInputParams)));
 
-            return (T)(await script.RunAsync(this.parameters).ConfigureAwait(false)).ReturnValue;
+            // Execute script and return the result.
+            // Parse Enum types explicitly since they cannot be casted directly.
+            var temp = (await script.RunAsync(this.parameters).ConfigureAwait(false)).ReturnValue;
+            return typeof(T).IsEnum ? (T)Enum.Parse(typeof(T), temp.ToString()) : (T)temp;
```
zhengmu, line 116: >(T)Enum.Parse [](start = 38, length = 13)

better to use Enum.IsDefined PLUS Enum.TryParse. see below.

Colors colorValue;
if (Enum.TryParse(colorString, true, out colorValue))
if (Enum.IsDefined(typeof(Colors), colorValue) |


-------------below example show why need both------------------
enum Colors
{
Red = 0,
Green = 1,
};


void Main()
{
 Console.WriteLine(Enum.Parse(typeof(Colors), "5"));
}
 
// output is 5,

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -923,6 +923,12 @@ public async Task<object> EvaluateDynamicProperty(dynamic schemaObj, Type knownT
                         }
                     }
 
+                    if (knownType?.IsEnum ?? false)
+                    {
+                        // Case when schema property is an Enum type.
+                        return Enum.Parse(knownType, schemaObj);
```
zhengmu, line 929: >Enum.Parse [](start = 31, length = 10)

same comment. need to verify with Enum.IsDefined

Forge.TreeWalker/src/ExpressionExecutor.cs
TravisOnGit: This is worth debating because I'm not sure to support non-defined integer values or not. I added this note in my PR description recognizing this hole.
"Note that integer (and integer as string) values that do not map to a defined Enum member are allowed. E.g. 100. This is because it is allowed in .Net and does not throw an exception. E.g. Console.WriteLine((FooEnum)100);."

---
In reply to: [426806360](https://github.com/microsoft/Forge/pull/15#discussion_r426806360) [](ancestors = 426806360)

zhengmu: 1) hope the string ""Note that integer (and integer as string) values that do not map..." is at line 114 or around, as future code reader (not the code review this time) is less likely to see the reason called out.

2) I am trying to see what is the use case scenario to have undefined Integer here, like 999. 

For new added enum value? It is possible but less likely, because Anvil has frequent deploy cycle. It should always have the latest Enum definition.

For a remove enum value? That is possible but we rarely do that. 

Do you see any scenario we shall allow undefined integer here? In Anvil, after parsing the string to Enum, the rest logic will use the Enum value, like Red/Green, not 0/1.

---
In reply to: [426827645](https://github.com/microsoft/Forge/pull/15#discussion_r426827645) [](ancestors = 426827645,426806360)

TravisOnGit: Discussed offline and decided it probably makes more sense to not add the additional safety checks.

* Argument to not add the safety checks is that undefined integer values are valid. And if you want the safety, you can use the string value. And we prefer to use strings in the schema in general since it is human readable. And there could be a hypothetical situation where you are using the schema as a pass-through, or the ForgeWrapper app deployment cycle is slower than endpoint that has latest Enum contract..
* Argument for adding safety checks is that it makes sense for Anvil case. We believe it would probably make sense in most cases.

---
In reply to: [426843656](https://github.com/microsoft/Forge/pull/15#discussion_r426843656) [](ancestors = 426843656,426827645,426806360)

PR #23, opened 2020-08-06
Adding option to construct a TreeWalkerParamemeters with a ForgeTree

TravisOnGit:
Adding option to construct a TreeWalkerParamemeters with a ForgeTree object directly, instead of the Json string. This saves the TreeWalkerSession from having to deserialize the json.

To take full advantage of this small feature, it is suggested for your application to cache the deserialized ForgeTree object(s), and use those when constructing TreeWalkerParameters.

### DIFF
--- a/Forge.TreeWalker/src/TreeWalkerParameters.cs
+++ b/Forge.TreeWalker/src/TreeWalkerParameters.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Forge.TreeWalker
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis.Scripting;
+    using Microsoft.Forge.DataContracts;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -117,7 +125,7 @@ public class TreeWalkerParameters
         #endregion
 
         /// <summary>
-        /// Instantiates a TreeWalkerParameters object with the properies that are required to instantiate a TreeWalkerSession object.
+        /// Instantiates a TreeWalkerParameters object with the properties that are required to instantiate a TreeWalkerSession object.
```
abhayketkar, line 128: >. [](start = 134, length = 1)

add comment mentioning one should prefer the other ctor to avoid deserializing each time. This ctor is here for backward compatibility

TravisOnGit: I'll put this in future PR, thanks.

PR #25, opened 2020-08-14
Major Roslyn Improvements to Memory and Running Time

TravisOnGit:
Roslyn Improvements Summary:
- **Memory** consumption down from 75MB per unique expression to a flat 25MB plus 0.3MB~ per unique expression. This is a 250x improvement or 0.4% memory consumed.
- **Running time** of first execution down from 2 seconds for each unique expression to a flat 2 seconds plus 15ms per unique expression. This is a 133x improvement or 0.75% running time.
- All changes are under the covers. No changes required by applications using Forge.

Feature breakdown:
- ParentScript
  - The parentScript that, upon initialization, gets Created, RunAsync, and added to the ScriptCache.
  - All Executed expressions are continued from the parentScript, avoiding additional compiles.
  - Before improvement - each unique expression would cost 25MB, take 2 seconds on first execution, then 0ms on further executions.
  - After improvement - only the parentScript costs 25MB and takes 2 seconds. Each unique expression costs < 0.5MB, takes 15ms on first execution, then 0ms on further executions.

- ParentScriptTask
  - The parentScriptTask kicks off initializing the Roslyn parentScript asynchronously, allowing ExpressionExecutor to construct very quickly.
  - This initialization time is saved if the application takes time to initialize before the first WalkTree call.
  - Before improvement - 1800ms for first initialize.
  - After improvement - 2ms for first Initialize. 0ms for further initializes (new TreeWalkerSessions using same ScriptCache).

- ParentScriptCode
  - Code that Forge uses to prime the parentScript on initialization.
  - Evaluating some code moves roughly 600ms from the first expression to asynchronously initializing the parentScript.

- ScriptCache rework
  - ScriptCache now caches Scripts that come from the parentScript.ContinueWith. ScriptCache still caches one Script per unique Expression, but these no longer get compiled.
  - ScriptCache now caches the ParentScript as well on initialization.

- MetadataReferenceResolver
  - This added ScriptOption causes Roslyn to not load missing references.
  - Before improvement - it costs about 75MB to initialize, then 0.5-1.5MB per unique expression.
  - After improvement -  it costs about 25MB to initialize, then 0.05-0.5MB per unique expression.

### DIFF
--- a/Forge.TreeWalker/src/ExpressionExecutor.cs
+++ b/Forge.TreeWalker/src/ExpressionExecutor.cs
@@ -26,7 +26,9 @@ namespace Microsoft.Forge.TreeWalker
     {
         /// <summary>
         /// Code that Forge uses to prime the parentScript on initialization.
-        /// Evaluating some code moves roughly 600ms from the first expression to initializing the parentScript.
+        /// Evaluating any random code moves roughly 600ms from the first expression to initializing the parentScript.
+        /// The first expression still takes roughly 100ms when tested.
+        /// The code block is arbitrary, though I did find in testing that the first expression ran quicker when the logic was similar to the parentScriptCode.
         /// </summary>
         public static string ParentScriptCode = "(1+1).ToString()";
 
@@ -51,8 +53,9 @@ namespace Microsoft.Forge.TreeWalker
         private List<Type> dependencies;
 
         /// <summary>
-        /// The ScriptCache is used to cache and re-use Roslyn scripts.
-        /// Scripts include the parentScript, as well as all unique expressions that get Executed.
+        /// The ScriptCache holds Roslyn Scripts that are created using parentScript.ContinueWith.
+        /// This saves on memory and time since these continued Scripts use the already compiled parentScript as a base.
+        /// The parentScript gets asynchronously compiled, ran, and cached on initialization.
         /// </summary>
         private ConcurrentDictionary<string, Script<object>> scriptCache;
 
--- a/Forge.TreeWalker/src/ExpressionExecutor.cs
+++ b/Forge.TreeWalker/src/ExpressionExecutor.cs
@@ -12,8 +12,10 @@ namespace Microsoft.Forge.TreeWalker
     using System;
     using System.Collections.Concurrent;
     using System.Collections.Generic;
+    using System.Collections.Immutable;
     using System.Reflection;
     using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp.Scripting;
     using Microsoft.CodeAnalysis.Scripting;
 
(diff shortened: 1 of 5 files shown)


Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -126,55 +128,57 @@ public class TreeWalkerParameters
...
             if (sessionId == Guid.Empty) throw new ArgumentNullException("sessionId");
-            if (string.IsNullOrWhiteSpace(jsonSchema)) throw new ArgumentNullException("jsonSchema");
+            if (forgeTree == null) throw new ArgumentNullException("forgeTree");
             if (forgeState == null) throw new ArgumentNullException("forgeState");
             if (callbacks == null) throw new ArgumentNullException("callbacks");
             if (token == null) throw new ArgumentNullException("token");
 
             this.SessionId = sessionId;
-            this.JsonSchema = jsonSchema;
+            this.ForgeTree = forgeTree;
```
zhengmu, line 152: just to confirm, looks like you just switched line 140-152 with line 169-181. Besides these, only some comments are changed, right?

Forge.TreeWalker/src/ExpressionExecutor.cs
```diff
@@ -23,19 +25,36 @@ namespace Microsoft.Forge.TreeWalker
     public class ExpressionExecutor
     {
         /// <summary>
-        /// List of external type dependencies needed to compile expressions.
+        /// Code that Forge uses to prime the parentScript on initialization.
+        /// Evaluating some code moves roughly 600ms from the first expression to initializing the parentScript.
         /// </summary>
-        private List<Type> dependencies;
+        public static string ParentScriptCode = "(1+1).ToString()";
```
zhengmu, line 31: >"(1+1).ToString()" [](start = 48, length = 18)

nit: consider to add the comment why to add the NO-OP statement.

Forge.TreeWalker/src/ExpressionExecutor.cs
```diff
@@ -23,19 +25,36 @@ namespace Microsoft.Forge.TreeWalker
...
+        /// <summary>
+        /// List of external type dependencies needed to compile expressions.
+        /// </summary>
+        private List<Type> dependencies;
 
         /// <summary>
-        /// Roslyn script options.
+        /// The ScriptCache is used to cache and re-use Roslyn scripts.
+        /// Scripts include the parentScript, as well as all unique expressions that get Executed.
         /// </summary>
```
zhengmu, line 56: ></summary> [](start = 12, length = 10)

consider to add your comment in the code review description:
ScriptCache now caches Scripts that come from the parentScript.ContinueWith. ScriptCache still caches one Script per unique Expression, but these no longer get compiled.

zhengmu:
>     /// </summary>

actually, consider to add your code review comments here, starting with "Roslyn Improvements Summary:...."

---
Refers to: Forge.TreeWalker/src/ExpressionExecutor.cs:24 in 9ed42b2. [](commit_id = 9ed42b248d8d3805ffec4cf1d3112ab2e9a08c59, deletion_comment = False)

Forge.TreeWalker/src/TreeWalkerParameters.cs
TravisOnGit: I swapped the constructors to put the ForgeTree one on top. Just because it is preferred to use that one. No functional changes.

---
In reply to: [470868342](https://github.com/microsoft/Forge/pull/25#discussion_r470868342) [](ancestors = 470868342)

Forge.TreeWalker/src/ExpressionExecutor.cs
TravisOnGit: I've outlined the various feature improvements on the new fields/properties. I describe more about the parentScript below. I didn't do deep investigation into how or why adding code here works. I just noticed it in the testing. It didn't matter too much what the code I put here was. I seemed to get slightly better times if the logic here matched the first executed expression logic, such as doing arithmetic. But since the improvements were small and it entirely depends on the unique ForgeTree that each application uses, I don't think its worth any more investment to investigate and improve.

---
In reply to: [470868792](https://github.com/microsoft/Forge/pull/25#discussion_r470868792) [](ancestors = 470868792)

Forge.TreeWalker/src/ExpressionExecutor.cs
TravisOnGit: Added more detail to ScriptCache description.

---
In reply to: [470870229](https://github.com/microsoft/Forge/pull/25#discussion_r470870229) [](ancestors = 470870229)

TravisOnGit:
>     /// </summary>

The history of this PR will not go away, so we can look back in the description to find the original notes if needed.
In general, I want the comments to reflect the current state of the code. I did include a brief note about the improvement numbers on parentScript, but I want to avoid adding too much historical info.

---
In reply to: [674277179](https://github.com/microsoft/Forge/pull/25#issuecomment-674277179) [](ancestors = 674277179)

---
Refers to: Forge.TreeWalker/src/ExpressionExecutor.cs:24 in 9ed42b2. [](commit_id = 9ed42b248d8d3805ffec4cf1d3112ab2e9a08c59, deletion_comment = False)

PR #32, opened 2020-10-13
WORK: add new Interface to support TreeNodeContext in Parameters.Call…

zhengmu:
…backs


Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -180,8 +188,89 @@ public class TreeWalkerParameters
...
+        #endregion Constructor with ITreeWalkerCallbacksV2
+
+        private void TryUpgradeToITreeWalkerCallbacksV2(ITreeWalkerCallbacks callbacks)
+        {
+            if (callbacks is ITreeWalkerCallbacksV2)
+            {
+                ITreeWalkerCallbacksV2 callbacksV2 = callbacks as ITreeWalkerCallbacksV2;
+                if (callbacksV2 != null)
+                {
+                    this.CallbacksV2 = callbacksV2;
```
TravisOnGit, line 268: >this.CallbacksV2 = callbacksV2; [](start = 20, length = 31)

Interesting idea to allow passing in ITreeWalkerCallbacks and attempting to convert to V2 on the users behalf. Since we are doing this, perhaps we don't even need the new constructors?

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,100 @@
...
+            dynamic properties,
+            object userContext,
+            CancellationToken token,
+            string treeName,
+            Guid rootSessionId,
+            bool shouldSkipActionsInTreeNode)
+        {
+            if (sessionId == null) throw new ArgumentNullException("sessionId");
+            if (string.IsNullOrWhiteSpace(treeNodeKey)) throw new ArgumentNullException("treeNodeKey");
+            if (userContext == null) throw new ArgumentNullException("userContext");
```
TravisOnGit, line 85: >userContext [](start = 16, length = 11)

userContext can be null. It's an optional parameter.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -305,31 +305,63 @@ public async Task<string> WalkTree(string treeNodeKey)
...
-                        this.Parameters.RootSessionId,
-                        this.walkTreeCts.Token).ConfigureAwait(false);
+                    TreeNodeContext treeNodeContext = null;
+                    if (this.Parameters.CallbacksV2 != null)
+                    {
+                        // If applicable, use the new CallbacksV2 with ITreeWalkerCallbacksV2
+                        // Evaluate the dynamic properties that are used by the actionTask.
+                        treeNodeContext = new TreeNodeContext(
+                            this.Parameters.SessionId,
+                            treeNodeKey,
```
TravisOnGit, line 315: >treeNodeKey [](start = 28, length = 11)

current instead of treeNodeKey

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -305,31 +305,63 @@ public async Task<string> WalkTree(string treeNodeKey)
...
-                        await this.EvaluateDynamicProperty(this.Schema.Tree[current].Properties, null),
-                        this.Parameters.UserContext,
-                        this.Parameters.TreeName,
-                        this.Parameters.RootSessionId,
-                        this.walkTreeCts.Token).ConfigureAwait(false);
+                    TreeNodeContext treeNodeContext = null;
+                    if (this.Parameters.CallbacksV2 != null)
+                    {
+                        // If applicable, use the new CallbacksV2 with ITreeWalkerCallbacksV2
+                        // Evaluate the dynamic properties that are used by the actionTask.
```
TravisOnGit, line 312: >// Evaluate the dynamic properties that are used by the actionTask. [](start = 24, length = 67)

Remove this comment.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -305,31 +305,63 @@ public async Task<string> WalkTree(string treeNodeKey)
...
+                        bool skipAction = treeNodeContext != null && treeNodeContext.ShouldSkipActionsInTreeNode;
+                        next = await this.VisitNode(current, skipAction).ConfigureAwait(false);
                     }
                     finally
                     {
-                        // Always call this callback, whether or not visit node was successful.
-                        await this.Parameters.Callbacks.AfterVisitNode(
+                        if (treeNodeContext != null)
+                        {
+                            // Follow the previous Callbacks with ITreeWalkerCallbacks.
```
TravisOnGit, line 350: >// Follow the previous Callbacks with ITreeWalkerCallbacks. [](start = 28, length = 59)

Comment does not line up with V2 scenario.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -305,31 +305,63 @@ public async Task<string> WalkTree(string treeNodeKey)
...
+                            await this.Parameters.CallbacksV2.AfterVisitNode(treeNodeContext).ConfigureAwait(false);
```
TravisOnGit, line 351: >treeNodeContext [](start = 77, length = 15)

Hmm.. Let's discuss re-using the same TreeNodeContext object for BeforeVisitNode and AfterVisitNode. It is the same object, so it makes sense to use the same object. However, the existing behavior recalculates the Properties value. The user could potentially be relying on Properties getting calculated twice. Or they could be assuming it should only be calculated once.. Not sure which way to go here.

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTestsWithV1Callbacks.cs
```diff
@@ -0,0 +1,1143 @@
+//-----------------------------------------------------------------------
+// <copyright file="TreeWalkerUnitTestsWithV1Callbacks.cs" company="Microsoft">
```
TravisOnGit, line 2: >TreeWalkerUnitTestsWithV1Callbacks [](start = 20, length = 34)

We should be able to test the new behavior in a few tests without having to duplicate the entire test file for this one change.

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -33,27 +33,57 @@ public class TreeWalkerUnitTests
...
+            TreeWalkerCallbacksV2 treeWalkerCallbacksV2 = new TreeWalkerCallbacksV2();
+            treeWalkerCallbacksV2.ShouldSkipActionsInTreeNode = shouldSkipActionsInTreeNode;
+            ITreeWalkerCallbacks callbacksV1 = treeWalkerCallbacksV2;
+            this.token = new CancellationTokenSource().Token;
+
+            this.parameters = new TreeWalkerParameters(
+                this.sessionId,
+                jsonSchema,
+                this.forgeState,
+                callbacksV1,            // Passes in ITreeWalkerCallbacks as V1, expects TreeWalkerParameters will recognize the instance is actually V2 and uses it as V2; otherwise throw NotImplementedException
```
TravisOnGit, line 86: >// Passes in ITreeWalkerCallbacks  [](start = 40, length = 34)

This comment is too long to be in-line. Please move it to its own line, possibly multi-line.

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -33,27 +33,57 @@ public class TreeWalkerUnitTests
...
-        public void TestInitialize(string jsonSchema, string treeName = null)
+        public void TestInitialize(string jsonSchema, string treeName = null, bool shouldSkipActionsInTreeNode = false)
         {
             // Initialize contexts, callbacks, and actions.
             this.sessionId = Guid.NewGuid();
             this.forgeState = new ForgeDictionary(new Dictionary<string, object>(), this.sessionId, this.sessionId);
-            this.callbacks = new TreeWalkerCallbacks();
+            TreeWalkerCallbacksV2 treeWalkerCallbacksV2 = new TreeWalkerCallbacksV2();
+            treeWalkerCallbacksV2.ShouldSkipActionsInTreeNode = shouldSkipActionsInTreeNode;
+            this.callbacksV2 = treeWalkerCallbacksV2;
```
TravisOnGit, line 51: I think a single line is cleaner here.
this.callbacksV2 = new TreeWalkerCallbacksV2() { ShouldSkipActionsInTreeNode: shouldSkipActionsInTreeNode};

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -406,43 +438,66 @@ public async Task<string> WalkTree(string treeNodeKey)
...
+                            break;
+                        }
                 }
             }
 
+            if (treeNode.Type == TreeNodeType.Leaf)
+            {
+                // Leaf type can't have ChildSelector so we return here.
+                return null;
+            }
```
TravisOnGit, line 499: Why did you move this?

Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -52,9 +52,16 @@ public class TreeWalkerParameters
 
         /// <summary>
```
TravisOnGit, line 53: >/// <summary> [](start = 8, length = 13)

Add a line:
/// [DEPRECATED]

Forge.TreeWalker/Forge.TreeWalker.csproj
```diff
@@ -58,7 +58,7 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
     <PackageReference Include="System.Collections.Immutable" Version="1.2.0" />
-    <PackageReference Include="System.Reflection.Metadata" Version="1.2.0" >
+    <PackageReference Include="System.Reflection.Metadata" Version="1.2.0">
```
TravisOnGit, line 61: >"> [](start = 73, length = 2)

nit: add space back.

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// Instantiates an TreeNodeContext object.
+        /// </summary>
+        /// <param name="sessionId">The unique identifier for this tree walking session.</param>
+        /// <param name="treeNodeKey">The TreeNode's key where the Action is taking place.</param>
+        /// <param name="properties">The properties of this Tree node..</param>
+        /// <param name="userContext">The user context for this Action.</param>
+        /// <param name="token">The cancellation token.</param>
+        /// <param name="treeName">The name of the ForgeTree in the JsonSchema.</param>
+        /// <param name="rootSessionId">The unique identifier for the root/parent tree walking session.</param>
+        /// <param name="ShouldSkipActionsInTreeNode">Indicates whether the tree walker should skip all actions defined in this node, and proceed with ChildSelector</param>
```
TravisOnGit, line 72: >r [](start = 163, length = 1)

nit: add period

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// Indicates whether the tree walker should skip all actions defined in this node, and proceed with ChildSelector.
+        /// </summary>
+        /// <remarks>Allow Public set for testing purpose.</remarks>
+        public bool ShouldSkipActionsInTreeNode { get; set; }
+
+        /// <summary>
+        /// Instantiates an TreeNodeContext object.
+        /// </summary>
+        /// <param name="sessionId">The unique identifier for this tree walking session.</param>
+        /// <param name="treeNodeKey">The TreeNode's key where the Action is taking place.</param>
```
TravisOnGit, line 66: >where the Action is taking place [](start = 57, length = 32)

This pasted wording from ActionContext doesn't make sense in TreeNodeContext.

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,100 @@
...
+// <summary>
+//     The TreeNodeContext class.
+// </summary>
+//-----------------------------------------------------------------------
+
+namespace Microsoft.Forge.TreeWalker
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
```
TravisOnGit, line 14: Can we remove these? I don't see any Tasks in this file.

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <summary>
+        /// The unique identifier for the root/parent tree walking session.
+        /// </summary>
+        public Guid RootSessionId { get; private set; }
+
+        /// <summary>
+        /// Indicates whether the tree walker should skip all actions defined in this node, and proceed with ChildSelector.
+        /// </summary>
+        /// <remarks>Allow Public set for testing purpose.</remarks>
+        public bool ShouldSkipActionsInTreeNode { get; set; }
```
TravisOnGit, line 60: We actually would rely on the application to update this property. It will be decided inside BeforeVisitNode if we should skip the Actions.

Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -180,8 +188,89 @@ public class TreeWalkerParameters
...
+
+            this.SessionId = sessionId;
+            this.JsonSchema = jsonSchema;
+            this.ForgeState = forgeState;
+            this.CallbacksV2 = callbacksV2;
+            this.Token = token;
+        }
+        #endregion Constructor with ITreeWalkerCallbacksV2
+
+        private void TryUpgradeToITreeWalkerCallbacksV2(ITreeWalkerCallbacks callbacks)
```
TravisOnGit, line 261: >TryUpgradeToITreeWalkerCallbacksV2 [](start = 21, length = 34)

Add missing summary.

Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -180,8 +188,89 @@ public class TreeWalkerParameters
...
+            {
+                ITreeWalkerCallbacksV2 callbacksV2 = callbacks as ITreeWalkerCallbacksV2;
+                if (callbacksV2 != null)
+                {
+                    this.CallbacksV2 = callbacksV2;
+                    return;
+                }
+            }
+
+            this.Callbacks = callbacks;
```
TravisOnGit, line 273: >this.Callbacks = callbacks [](start = 12, length = 26)

Here's another idea for implementing the V2 backwards compatibility logic:

Keep only the public property ITreeWalkerCallbacks.
Have a private bool that is true if Callbacks is castable to V2.
As an optimization to save ourselves from casting Callbacks to V2 each time, also cache a private ITreeWalkerCallbacksV2 property here, and use that when bool is true.
Retrieve these new private values either through Get methods or perhaps just make them public with private set.

The benefits of doing it this way:
No additional public properties to worry about.
No additional constructors to worry about.
V2 is automatically taken care of under the covers.

Forge.TreeWalker/src/TreeWalkerSession.cs
TravisOnGit: Another thing to consider is that BeforeVisitNode will be manipulating property values of TreeNodeContext, such as the new SkipAction property. Given that case, we would need to pass the same object to AfterVisitNode.

---
In reply to: [504167224](https://github.com/microsoft/Forge/pull/32#discussion_r504167224) [](ancestors = 504167224)

TravisOnGit:
>                         await this.CommitActionResponse(treeActionKey, timeoutResponse).ConfigureAwait(false);

I was thinking if SkipAction property is true, how does the schema know that happened? Schema must know this happened so it can respond accordingly in the ChildSelector. Here are a few ideas

Option 1, persist new state.
    ShouldSelect: C#|Session.SkippedCurrentTreeNode, Child: FallbackChildFoo

Option 2, set the ActionResponse with Skipped Status and check for that.
    ShouldSelect: C#|Session.GetLastActionResponse().Status == "Skipped", Child: FallbackChildFoo

Option3, OnBeforeVisitNode has new property specifying the desired Child to go to. Maybe this gets persisted?
    ShouldSelect: C#|Session.FallbackChild == "Foo", Child: Foo

---
Refers to: Forge.TreeWalker/src/TreeWalkerSession.cs:800 in 991a747. [](commit_id = 991a74737461f4ed1c80b41b1456fde6940e12ff, deletion_comment = False)

Forge.TreeWalker/Forge.TreeWalker.csproj
zhengmu: The visual studio 2019 auto removed this space. After double check, looks like the space actually should be removed because this is a complete XML element (<A></A>), not a compact one (<A />)

---
In reply to: [504219758](https://github.com/microsoft/Forge/pull/32#discussion_r504219758) [](ancestors = 504219758)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: The visual studio 2019 auto removed this space. After double check, looks like the space actually should be removed because this is a complete XML element (<A></A>), not a compact one (<A />)

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: fixed

---
In reply to: [504219343](https://github.com/microsoft/Forge/pull/32#discussion_r504219343) [](ancestors = 504219343)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: fixed

---
In reply to: [504220866](https://github.com/microsoft/Forge/pull/32#discussion_r504220866) [](ancestors = 504220866)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: yes, only application and test cases should update this property. i added your comments to inline.

---
In reply to: [504223272](https://github.com/microsoft/Forge/pull/32#discussion_r504223272) [](ancestors = 504223272)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: good catch. removed.

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: fixed

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: Technically, yes, with this change, we don't need the new constructors. However, we are encouraging users to use the new constructors if possible. One we confirm there is no user for the old constructor, we can remove them.

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: Yes, you list another option to do it:

No additional public properties to worry about. // yes, no additional public. but need a new Get function to get the V2.

No additional constructors to worry about.  // yes, users can keep use the existing constructor without breaking anything. On the second thought, our purpose is to let them switch to the new interface as early as possible. If they keep use the V1 interface, then we have to keep the if-else logic in TreeWalkerSession.WalkTree(), which is we want to avoid. Leave the V1 interface is kind of encourage the opposite in the long run.

V2 is automatically taken care of under the covers. // Same reason as above. I would like to convert all users to the new V2 Interface and only support V1 under the covers as a fallback plan.

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: fixed

---
In reply to: [504223719](https://github.com/microsoft/Forge/pull/32#discussion_r504223719) [](ancestors = 504223719)

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: Yes, you list another option to do it:

No additional public properties to worry about. // yes, no additional public. but need a new Get function to get the V2.

No additional constructors to worry about.  // yes, users can keep use the existing constructor without breaking anything. On the second thought, our purpose is to let them switch to the new interface as early as possible. If they keep use the V1 interface, then we have to keep the if-else logic in TreeWalkerSession.WalkTree(), which is we want to avoid. Leave the V1 interface is kind of encourage the opposite in the long run.

V2 is automatically taken care of under the covers. // Same reason as above. I would like to convert all users to the new V2 Interface and only support V1 under the covers as a fallback plan.

---
In reply to: [504228085](https://github.com/microsoft/Forge/pull/32#discussion_r504228085) [](ancestors = 504228085)

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: Technically, yes, with this change, we don't need the new constructors. However, we are encouraging users to use the new constructors if possible. One we confirm there is no user for the old constructor, we can remove them.

---
In reply to: [504157718](https://github.com/microsoft/Forge/pull/32#discussion_r504157718) [](ancestors = 504157718)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: good catch. removed.

---
In reply to: [504221418](https://github.com/microsoft/Forge/pull/32#discussion_r504221418) [](ancestors = 504221418)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: removed the null check, and added the comment
userContext is an optional parameter for TreeNode, so we don't throw ArgumentNullException when null.

---
In reply to: [504159484](https://github.com/microsoft/Forge/pull/32#discussion_r504159484) [](ancestors = 504159484)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: Good catch! Pretty sharp eye in code view.

---
In reply to: [504221641](https://github.com/microsoft/Forge/pull/32#discussion_r504221641) [](ancestors = 504221641)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: removed

---
In reply to: [504163370](https://github.com/microsoft/Forge/pull/32#discussion_r504163370) [](ancestors = 504163370)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: wow, this is a good bug! thanks

---
In reply to: [504162909](https://github.com/microsoft/Forge/pull/32#discussion_r504162909) [](ancestors = 504162909)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: fixed the comment.

---
In reply to: [504165345](https://github.com/microsoft/Forge/pull/32#discussion_r504165345) [](ancestors = 504165345)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: agree with you on both. So i now changed it to use different objects, for the recalculating, and copy the value of ShouldSkipActionsInTreeNode

---
In reply to: [504228812](https://github.com/microsoft/Forge/pull/32#discussion_r504228812) [](ancestors = 504228812,504167224)

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -406,43 +438,66 @@ public async Task<string> WalkTree(string treeNodeKey)
...
         /// <param name="treeNodeKey">The TreeNode key to visit.</param>
+        /// <param name="skipAction">Specifies whether to skip all actions and do ChildSelector only.</param>
         /// <exception cref="TimeoutException">If the node-level timeout was hit.</exception>
         /// <exception cref="ActionTimeoutException">If the action-level timeout was hit.</exception>
         /// <exception cref="OperationCanceledException">If the cancellation token was triggered.</exception>
         /// <exception cref="Exception">If an unexpected exception was thrown.</exception>
         /// <returns>The key of the next child to visit, or <c>null</c> if no match was found.</returns>
         public async Task<string> VisitNode(string treeNodeKey)
+        {
+            return await this.VisitNode(treeNodeKey, skipAction: false);
```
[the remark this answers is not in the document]
zhengmu: good tips, and it can save one more context switch by saving one async await.

---
In reply to: [504169506](https://github.com/microsoft/Forge/pull/32#discussion_r504169506) [](ancestors = 504169506)

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -406,43 +438,66 @@ public async Task<string> WalkTree(string treeNodeKey)
         /// Visits a TreeNode in the ForgeTree, performing type-specific behavior as necessary before selecting the next child to visit.
         /// </summary>
         /// <param name="treeNodeKey">The TreeNode key to visit.</param>
+        /// <param name="skipAction">Specifies whether to skip all actions and do ChildSelector only.</param>
```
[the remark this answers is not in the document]
zhengmu: removed

---
In reply to: [504169707](https://github.com/microsoft/Forge/pull/32#discussion_r504169707) [](ancestors = 504169707)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: because regardless skipAction is true or false, we need the check below. So i moved the code to the bottom.

            if (treeNode.Type == TreeNodeType.Leaf)
            {
                // Leaf type can't have ChildSelector so we return here.
                return null;
            }

---
In reply to: [504218989](https://github.com/microsoft/Forge/pull/32#discussion_r504218989) [](ancestors = 504218989)

zhengmu:
>                         await this.CommitActionResponse(treeActionKey, timeoutResponse).ConfigureAwait(false);

Option 2, we need to commit the LastAction while all actions in this node are skipped. not sure how to do it.

Option 1, we need to set Session.SkippedCurrentTreeNode to true before selection, and set the value to false after selection.

Option 3, limit the FallbackChild to only one node, while we should allow multiple choices.

let me try option 1

---
In reply to: [707985867](https://github.com/microsoft/Forge/pull/32#issuecomment-707985867) [](ancestors = 707985867)

---
Refers to: Forge.TreeWalker/src/TreeWalkerSession.cs:800 in 991a747. [](commit_id = 991a74737461f4ed1c80b41b1456fde6940e12ff, deletion_comment = False)

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTestsWithV1Callbacks.cs
zhengmu: yes, we should be able to update a few tests. My thought is to use a copied file to cover all existing tests with V1, so V1 is fully verified. Then update the existing tests with V2, so V2 are fully covered. And we can delete "TreeWalkerUnitTestsWithV1Callbacks.cs" in future when we officially retire V1 interface.

---
In reply to: [504215906](https://github.com/microsoft/Forge/pull/32#discussion_r504215906) [](ancestors = 504215906)

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
zhengmu: fixed

---
In reply to: [504217378](https://github.com/microsoft/Forge/pull/32#discussion_r504217378) [](ancestors = 504217378)

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
zhengmu: totally agree, changed to 

this.callbacksV2 = new TreeWalkerCallbacksV2() { ShouldSkipActionsInTreeNode = shouldSkipActionsInTreeNode };

---
In reply to: [504218674](https://github.com/microsoft/Forge/pull/32#discussion_r504218674) [](ancestors = 504218674)

Forge.TreeWalker/src/TreeWalkerParameters.cs
TravisOnGit: We cannot know when users switch to V2 because we can't tell which version folks are using. The transition will naturally happen over time when folks want to use new features that are in V2.
I'm not sure which implementation to go with, they all seem fine. We want the implementation to be as smooth as possible when we introduce V2, when we eventually deprecate V1, and with least duplicate code.

---
In reply to: [504278218](https://github.com/microsoft/Forge/pull/32#discussion_r504278218) [](ancestors = 504278218,504228085)

Forge.TreeWalker/src/TreeNodeContext.cs
TravisOnGit: "
When not null or whitespace, the tree walker will skip all actions defined in the current tree node, and proceed to AfterVisitNode then ChildSelector.
Update this property inside BeforeVisitNode if you wish to use this feature for the current tree node.
The string context is available to check in the current TreeNode's ChildSelector via Session.GetCurrentNodeSkipActionContext().
"

---
In reply to: [504257524](https://github.com/microsoft/Forge/pull/32#discussion_r504257524) [](ancestors = 504257524,504223272)

zhengmu: updated the comment

---
In reply to: [505093484](https://github.com/microsoft/Forge/pull/32#discussion_r505093484) [](ancestors = 505093484,504257524,504223272)

TravisOnGit: Please replace the entire summary with what I have above.

---
In reply to: [505150928](https://github.com/microsoft/Forge/pull/32#discussion_r505150928) [](ancestors = 505150928,505093484,504257524,504223272)

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,105 @@
...
+            Guid sessionId,
+            string treeNodeKey,
+            dynamic properties,
+            object userContext,
+            CancellationToken token,
+            string treeName,
+            Guid rootSessionId,
+            string currentNodeSkipActionContext)
+        {
+            // userContext is an optional parameter for TreeNode, so we don't throw ArgumentNullException when null.
```
TravisOnGit, line 87: >// userContext is an optional parameter for TreeNode, so we don't throw ArgumentNullException when null. [](start = 12, length = 104)

Please remove this comment. UserContext being an optional parameter is established in TreeWalkerParameters as well as on the wiki. We don't need to remind folks in more places of this fact.

Forge.TreeWalker/src/TreeWalkerParameters.cs
TravisOnGit: callbacks param is empty.

---
In reply to: [504277996](https://github.com/microsoft/Forge/pull/32#discussion_r504277996) [](ancestors = 504277996,504223719)

Forge.TreeWalker/src/TreeWalkerParameters.cs
TravisOnGit: I agree with you here about having the V2 constructors. We want folks to use them so eventually we can remove the V1 constructors and interface.

Because of this, let's force folks to use the V2 constructor if they wish to take advantage of the V2 features. We do this by removing TryUpgradeToITreeWalkerCallbacksV2. Folks who don't upgrade continue to use V1 interface and constructor. Folks who upgrade use the V2 interface and constructor.

V1 and V2 are now completely separate. I think this allows us to remove the inheritance of V1 from V2 and treat them completely separately. This has the benefit of V2 users not having to pseudo-implement the V1 interface. TreeWalkerSession code already works like this as well. Let me know what you think.

---
In reply to: [504869930](https://github.com/microsoft/Forge/pull/32#discussion_r504869930) [](ancestors = 504869930,504278218,504228085)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: sure, will remove

---
In reply to: [505812830](https://github.com/microsoft/Forge/pull/32#discussion_r505812830) [](ancestors = 505812830)

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -305,14 +319,37 @@ public async Task<string> WalkTree(string treeNodeKey)
...
+                            this.Parameters.UserContext,
+                            this.walkTreeCts.Token,
+                            this.Parameters.TreeName,
+                            this.Parameters.RootSessionId,
+                            currentNodeSkipActionContext:null
+                        );
+
+                        // Follow the previous Callbacks with ITreeWalkerCallbacks.
+                        await this.Parameters.CallbacksV2.BeforeVisitNode(treeNodeContext_BeforeVisitNode).ConfigureAwait(false);
+                        this.currentNodeSkipActionContext = treeNodeContext_BeforeVisitNode?.CurrentNodeSkipActionContext;
```
TravisOnGit, line 339: >treeNodeContext_BeforeVisitNode? [](start = 60, length = 32)

Are you null checking here in case BeforeVisitNode sets TreeNodeContext to null for some reason? This may be overly safe, but I'm okay with it.

Forge.TreeWalker/src/TreeWalkerSession.cs
TravisOnGit: Ok sounds good. Doesn't look like you set currentNodeSkipActionContext on AfterVisitNode's TreeNodeContext. This should be set to the value from BeforeVistNode's TreeNodeContext.

---
In reply to: [504291737](https://github.com/microsoft/Forge/pull/32#discussion_r504291737) [](ancestors = 504291737,504228812,504167224)

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: cool. i like the idea to separate the two interfaces completely by removing TryUpgradeToITreeWalkerCallbacksV2().  Let me remove it and the associated test case.

---
In reply to: [505818977](https://github.com/microsoft/Forge/pull/32#discussion_r505818977) [](ancestors = 505818977,504869930,504278218,504228085)

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -321,18 +358,39 @@ public async Task<string> WalkTree(string treeNodeKey)
                     }
                     finally
                     {
-                        // Always call this callback, whether or not visit node was successful.
-                        await this.Parameters.Callbacks.AfterVisitNode(
+                        if (treeNodeContext_BeforeVisitNode != null)
```
TravisOnGit, line 361: >if (treeNodeContext_BeforeVisitNode != null) [](start = 24, length = 44)

If we're just using treeNodeContext_BeforeVisitNode for this null check, I'd prefer to just use the same check here:
if (this.Parameters.CallbacksV2 != null)

Then we don't have to use 2 different names for TreeNodeContext variable. Since they will both live inside separate if statements, we can call them both treeNodeContext.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -321,18 +358,39 @@ public async Task<string> WalkTree(string treeNodeKey)
...
                             await this.EvaluateDynamicProperty(this.Schema.Tree[current].Properties, null),
                             this.Parameters.UserContext,
                             this.Parameters.TreeName,
                             this.Parameters.RootSessionId,
                             this.walkTreeCts.Token).ConfigureAwait(false);
+                        }
                     }
 
                     current = next;
+                    this.currentNodeSkipActionContext = null;   // Clear the context of SkipAction and get ready for the next tree node.
```
TravisOnGit, line 392: >this.currentNodeSkipActionContext = null; [](start = 20, length = 41)

Let's move this inside the finally block. We want to make sure this gets cleared.

Forge.TreeWalker/src/TreeWalkerSession.cs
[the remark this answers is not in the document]
TravisOnGit: I still see the SkipAction param in iteration 5.

---
In reply to: [504295201](https://github.com/microsoft/Forge/pull/32#discussion_r504295201) [](ancestors = 504295201,504169707)

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -415,34 +474,43 @@ public async Task<string> VisitNode(string treeNodeKey)
...
             TreeNode treeNode = this.Schema.Tree[treeNodeKey];
 
-            // Do type-specific behavior.
-            switch (treeNode.Type)
+            if (string.IsNullOrWhiteSpace(this.currentNodeSkipActionContext))
             {
-                case TreeNodeType.Leaf:
-                {
-                    await this.PerformLeafTypeBehavior(treeNode).ConfigureAwait(false);
+                // When this.currentNodeSkipActionContext, it means there is no SkipAction. Just follow the regular route to run Actions.
```
TravisOnGit, line 479: >// When this.currentNodeSkipActionContext, it means there is no SkipAction. Just follow the regular route to run Actions. [](start = 16, length = 121)

Do not skip actions when this.currentNodeSkipActionContext is null or whitespace.

Forge.TreeWalker/src/TreeNodeContext.cs
TravisOnGit: "When set, the tree walker..."
I like that better.

---
In reply to: [505811771](https://github.com/microsoft/Forge/pull/32#discussion_r505811771) [](ancestors = 505811771,505150928,505093484,504257524,504223272)

Forge.TreeWalker/src/ITreeSession.cs
```diff
@@ -41,5 +41,11 @@ public interface ITreeSession
         /// </summary>
         /// <returns>The ActionResponse data for the last executed tree action key if it exists, otherwise null.</returns>
         Task<ActionResponse> GetLastActionResponseAsync();
+
+        /// <summary>
+        /// Get the string context if the actions in this TreeNodes are skipped.
+        /// Usually, when there is no action skipped, the return value is null.
```
TravisOnGit, line 47: Gets the string context if the actions in the current TreeNode were skipped, or null if actions were not skipped.

TravisOnGit:
>     /// The ITreeSession interface holds accessor methods into the forgeState dictionary.

and local TreeWalkerSession state.

---
Refers to: Forge.TreeWalker/src/ITreeSession.cs:15 in 114f4cc. [](commit_id = 114f4cce9845d741b10d6cc01a669bd3c2616382, deletion_comment = False)

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// </summary>
+        /// <remarks>Allow Public set for testing purpose.</remarks>
+        public string CurrentNodeSkipActionContext { get; set; }
+
+        /// <summary>
+        /// Instantiates an TreeNodeContext object.
+        /// </summary>
+        /// <param name="sessionId">The unique identifier for this tree walking session.</param>
+        /// <param name="treeNodeKey">The TreeNode's key.</param>
+        /// <param name="properties">The properties of this Tree node..</param>
```
TravisOnGit, line 69: >Tree node.. [](start = 60, length = 11)

Remove double period.

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// <param name="userContext">Optional, the user context for this Tree node.</param>
```
TravisOnGit, line 70: >Tree node [](start = 74, length = 9)

Let's be consistent how we write out tree node / TreeNode / Tree node.

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: good catch. fixed

---
In reply to: [505826440](https://github.com/microsoft/Forge/pull/32#discussion_r505826440) [](ancestors = 505826440,504291737,504228812,504167224)

Forge.TreeWalker/src/TreeWalkerSession.cs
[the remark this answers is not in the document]
zhengmu: removed

---
In reply to: [505830690](https://github.com/microsoft/Forge/pull/32#discussion_r505830690) [](ancestors = 505830690,504295201,504169707)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: cool. fixed

---
In reply to: [505841676](https://github.com/microsoft/Forge/pull/32#discussion_r505841676) [](ancestors = 505841676,505811771,505150928,505093484,504257524,504223272)

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: As suggested in another comment, removed this function. We will have complete separate V1 and V2.

---
In reply to: [505813758](https://github.com/microsoft/Forge/pull/32#discussion_r505813758) [](ancestors = 505813758,504277996,504223719)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: good point, removed the "?"

---
In reply to: [505822703](https://github.com/microsoft/Forge/pull/32#discussion_r505822703) [](ancestors = 505822703)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: the TreeNodeContext for afterVisitNode, need to get the currentNodeSkipActionContext from the TreeNodeContext for beforeVisitNode. So we  need two variable names, unless we use a local string variable to save the context from beforeVisitNode. Let me know if you think else.

---
In reply to: [505828075](https://github.com/microsoft/Forge/pull/32#discussion_r505828075) [](ancestors = 505828075)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: fixed

---
In reply to: [505830289](https://github.com/microsoft/Forge/pull/32#discussion_r505830289) [](ancestors = 505830289)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: fixed

---
In reply to: [505832458](https://github.com/microsoft/Forge/pull/32#discussion_r505832458) [](ancestors = 505832458)

Forge.TreeWalker/src/ITreeSession.cs
zhengmu: fixed

---
In reply to: [505844640](https://github.com/microsoft/Forge/pull/32#discussion_r505844640) [](ancestors = 505844640)

zhengmu:
>     /// The ITreeSession interface holds accessor methods into the forgeState dictionary.

updated

---
In reply to: [709586963](https://github.com/microsoft/Forge/pull/32#issuecomment-709586963) [](ancestors = 709586963)

---
Refers to: Forge.TreeWalker/src/ITreeSession.cs:15 in 114f4cc. [](commit_id = 114f4cce9845d741b10d6cc01a669bd3c2616382, deletion_comment = False)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: fixed

---
In reply to: [505849122](https://github.com/microsoft/Forge/pull/32#discussion_r505849122) [](ancestors = 505849122)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: choose "tree node"

---
In reply to: [505850342](https://github.com/microsoft/Forge/pull/32#discussion_r505850342) [](ancestors = 505850342)

Forge.TreeWalker/src/TreeNodeContext.cs
TravisOnGit: Can you remove the remarks? It isn't public just for testing.

---
In reply to: [505857375](https://github.com/microsoft/Forge/pull/32#discussion_r505857375) [](ancestors = 505857375,505841676,505811771,505150928,505093484,504257524,504223272)

Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -180,8 +192,77 @@ public class TreeWalkerParameters
...
+            if (forgeState == null) throw new ArgumentNullException("forgeState");
+            if (callbacksV2 == null) throw new ArgumentNullException("callbacksV2");
+            if (token == null) throw new ArgumentNullException("token");
+
+            this.SessionId = sessionId;
+            this.JsonSchema = jsonSchema;
+            this.ForgeState = forgeState;
+            this.CallbacksV2 = callbacksV2;
             this.Token = token;
         }
```
TravisOnGit, line 264: Let's remove this constructor. I want folks to use ForgeTree instead of jsonSchema. Introducing a new V2 interface gives a good opportunity to phase jsonSchema out as well.

Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -51,10 +51,18 @@ public class TreeWalkerParameters
         public IForgeDictionary ForgeState { get; private set; }
 
         /// <summary>
+        /// [DEPRECATED]
         /// The ITreeWalkerCallbacks interface defines the callback Tasks that are awaited while walking the tree.
+        /// Prefers future callers to use the new ITreeWalkerCallbacksV2 CallbacksV2, and keep this ITreeWalkerCallbacks Callbacks for backward compatibility.
```
TravisOnGit, line 56: >, and keep this ITreeWalkerCallbacks Callbacks for backward compatibility [](start = 84, length = 73)

Remove this part.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -321,18 +358,40 @@ public async Task<string> WalkTree(string treeNodeKey)
                     }
                     finally
                     {
-                        // Always call this callback, whether or not visit node was successful.
-                        await this.Parameters.Callbacks.AfterVisitNode(
+                        if (this.Parameters.CallbacksV2 != null)
+                        {
+                            // When CallbacksV2 is not null, continues with the new ITreeWalkerCallbacksV2.
```
TravisOnGit, line 363: >// When CallbacksV2 is not null, continues with the new ITreeWalkerCallbacksV2. [](start = 28, length = 79)

Let's use the same comment in BeforeVisitNode and AfterVisitNode.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -321,18 +358,40 @@ public async Task<string> WalkTree(string treeNodeKey)
...
+                            treeNodeContext = new TreeNodeContext(
```
TravisOnGit, line 364: >treeNodeContext = new TreeNodeContext( [](start = 28, length = 38)

Let's add a comment here since its not obvious why we are recreating the TreeNodeContext.

// Recreating the TreeNodeContext here to reevaluate TreeNode.Properties.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -270,6 +275,15 @@ public async Task<string> GetLastTreeAction()
             }
         }
 
+        /// <summary>
+        /// Get the string context if the actions in this TreeNodes are skipped.
+        /// Usually, when there is no action skipped, the return value is null.
+        /// </summary>
+        public string GetCurrentNodeSkipActionContext()
```
TravisOnGit, line 282: >GetCurrentNodeSkipActionContext [](start = 22, length = 31)

Update summary to match interface summary.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -138,6 +138,11 @@ public class TreeWalkerSession : ITreeSession
         /// </summary>
         private bool hasSessionRehydrated;
 
+        /// <summary>
+        /// The context value whether the actions in this TreeNodes are skipped.
+        /// </summary>
+        private string currentNodeSkipActionContext;
```
TravisOnGit, line 144: >currentNodeSkipActionContext [](start = 23, length = 28)

Let's use a similar summary as the get method.

/// The string context if the actions in the current TreeNode were skipped, or null if actions were not skipped.

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -33,27 +33,58 @@ public class TreeWalkerUnitTests
 
         private Guid sessionId;
         private IForgeDictionary forgeState = new ForgeDictionary(new Dictionary<string, object>(), Guid.Empty, Guid.Empty);
-        private ITreeWalkerCallbacks callbacks;
+        private ITreeWalkerCallbacksV2 callbacksV2;             // 20201012, switch to V2 callback.
```
TravisOnGit, line 36: >             // 20201012, switch to V2 callback. [](start = 51, length = 48)

Please remove this comment. I don't think timestamped comments when code was changed are helpful.

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -153,6 +184,19 @@ public void TestTreeWalkerSession_VisitNode_Success()
             Assert.AreEqual(expected, actualNextTreeNodeKey, "Expected VisitNode(Container) to return Tardigrade.");
         }
 
+        [TestMethod]
+        public void TestTreeWalkerSession_VisitNode_Success_WithV1Callbacks()
```
TravisOnGit, line 188: >TestTreeWalkerSession_VisitNode_Success_WithV1Callbacks [](start = 20, length = 55)

This test calls VisitNode directly, which avoids use of the ITreeWalkerCallbacks completely. Instead, copy the TestTreeWalkerSession_WalkTree_Success test.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -305,14 +319,37 @@ public async Task<string> WalkTree(string treeNodeKey)
...
+                            current,
+                            await this.EvaluateDynamicProperty(this.Schema.Tree[current].Properties, null),
+                            this.Parameters.UserContext,
+                            this.walkTreeCts.Token,
+                            this.Parameters.TreeName,
+                            this.Parameters.RootSessionId,
+                            currentNodeSkipActionContext:null
+                        );
+
+                        // Follow the previous Callbacks with ITreeWalkerCallbacks.
```
TravisOnGit, line 337: >// Follow the previous Callbacks with ITreeWalkerCallbacks. [](start = 24, length = 59)

Remove this duplicate and misplaced comment.

Forge.TreeWalker/src/TreeWalkerSession.cs
TravisOnGit: "Use CallbacksV2 when available."

---
In reply to: [505912854](https://github.com/microsoft/Forge/pull/32#discussion_r505912854) [](ancestors = 505912854)

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -321,18 +358,40 @@ public async Task<string> WalkTree(string treeNodeKey)
...
+                                this.Parameters.TreeName,
+                                this.Parameters.RootSessionId,
+                                this.currentNodeSkipActionContext
+                            );
+
+                            await this.Parameters.CallbacksV2.AfterVisitNode(treeNodeContext).ConfigureAwait(false);
+                        }
+                        else
+                        {
+                            // Always call this callback, whether or not visit node was successful.
```
TravisOnGit, line 379: >// Always call this callback, whether or not visit node was successful. [](start = 28, length = 71)

Let's move this comment so it is valid for V2 or V1. Perhaps above the if statement.

// Always call AfterVisitNode, even if VisitNode threw exception.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -321,18 +358,40 @@ public async Task<string> WalkTree(string treeNodeKey)
...
                             this.Parameters.SessionId,
                             current,
                             await this.EvaluateDynamicProperty(this.Schema.Tree[current].Properties, null),
                             this.Parameters.UserContext,
                             this.Parameters.TreeName,
                             this.Parameters.RootSessionId,
                             this.walkTreeCts.Token).ConfigureAwait(false);
+                        }
+
+                        this.currentNodeSkipActionContext = null;   // Clear the context of SkipAction and get ready for the next tree node.
```
TravisOnGit, line 390: >// Clear the context of SkipAction and get ready for the next tree node. [](start = 68, length = 72)

Let's keep comments on their own lines.

TravisOnGit: // Clear the context of SkipAction before visiting next node because it is only valid locally for this current tree node.

---
In reply to: [505919008](https://github.com/microsoft/Forge/pull/32#discussion_r505919008) [](ancestors = 505919008)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: fixed

---
In reply to: [505917480](https://github.com/microsoft/Forge/pull/32#discussion_r505917480) [](ancestors = 505917480)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: make sense. fixed

---
In reply to: [505918812](https://github.com/microsoft/Forge/pull/32#discussion_r505918812) [](ancestors = 505918812)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: replaced with your line.

---
In reply to: [505919350](https://github.com/microsoft/Forge/pull/32#discussion_r505919350) [](ancestors = 505919350,505919008)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: yes, in the latest design. it is not for testing only. Removed

---
In reply to: [505910307](https://github.com/microsoft/Forge/pull/32#discussion_r505910307) [](ancestors = 505910307,505857375,505841676,505811771,505150928,505093484,504257524,504223272)

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: i see. removed.

---
In reply to: [505911437](https://github.com/microsoft/Forge/pull/32#discussion_r505911437) [](ancestors = 505911437)

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: removed

---
In reply to: [505911809](https://github.com/microsoft/Forge/pull/32#discussion_r505911809) [](ancestors = 505911809)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: fixed => // If applicable, use the new CallbacksV2 with ITreeWalkerCallbacksV2

---
In reply to: [505917661](https://github.com/microsoft/Forge/pull/32#discussion_r505917661) [](ancestors = 505917661,505912854)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: i was thinking the same before iteration #5, but forgot to add it. Fixed.

---
In reply to: [505913873](https://github.com/microsoft/Forge/pull/32#discussion_r505913873) [](ancestors = 505913873)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: updated, also changed "TreeNode" to "tree node" for consistency. 
and updated the comment for private string currentNodeSkipActionContext;

---
In reply to: [505915013](https://github.com/microsoft/Forge/pull/32#discussion_r505915013) [](ancestors = 505915013)

Forge.TreeWalker/src/TreeWalkerSession.cs
zhengmu: updated

---
In reply to: [505915501](https://github.com/microsoft/Forge/pull/32#discussion_r505915501) [](ancestors = 505915501)

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
zhengmu: removed.

---
In reply to: [505915856](https://github.com/microsoft/Forge/pull/32#discussion_r505915856) [](ancestors = 505915856)

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
zhengmu: fixed. and updated a few test methods to cover V1 with File/ForgeTree, and V2 with ForgeTree

---
In reply to: [505916607](https://github.com/microsoft/Forge/pull/32#discussion_r505916607) [](ancestors = 505916607)

Forge.TreeWalker/src/TreeNodeContext.cs
TravisOnGit: /// <summary>
        /// When set, the tree walker will skip all actions defined in the current tree node, and proceed to AfterVisitNode then ChildSelector.
        /// Update this property inside BeforeVisitNode if you wish to use this feature for the current tree node.
        /// The string context is available to check in the current TreeNode's ChildSelector via Session.GetCurrentNodeSkipActionContext().
        /// </summary>

---
In reply to: [505924425](https://github.com/microsoft/Forge/pull/32#discussion_r505924425) [](ancestors = 505924425,505910307,505857375,505841676,505811771,505150928,505093484,504257524,504223272)

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -33,27 +33,60 @@ public class TreeWalkerUnitTests
...
+
+            this.session = new TreeWalkerSession(this.parameters);
+        }
+
+        /// <summary>
+        /// Helper function to test V1 ITreeWalkerCallbacks on the contructor accepting FileSchema.
+        /// </summary>
+        /// <param name="jsonSchema">The json string content.</param>
+        /// <param name="treeName">The tree name.</param>
+        public void TestInitializeWithFile_WithV1Callbacks(string jsonSchema, string treeName = null)
```
TravisOnGit, line 77: >TestInitializeWithFile [](start = 20, length = 22)

This isn't initializing with filePath. We can call it TestInitializeWithJsonSchema_WithV1Callbacks

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -67,19 +100,49 @@ public void TestInitialize(string jsonSchema, string treeName = null)
...
+                ForgeActionsAssembly = typeof(CollectDiagnosticsAction).Assembly,
+                InitializeSubroutineTree = this.InitializeSubroutineTree,
+                TreeName = treeName,
+                Dependencies = new List<Type>() { typeof(FooEnum) }
+            };
+
+            this.session = new TreeWalkerSession(this.parameters);
+        }
+
         public void TestInitializeWithForgeTree(ForgeTree forgeTree, string treeName = null)
```
TravisOnGit, line 133: >public void TestInitializeWithForgeTree [](start = 8, length = 39)

Let's move this initializer just under TestInitialize(..) so they are next to each other. The V1 initializers can follow.

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -33,27 +33,60 @@ public class TreeWalkerUnitTests
...
+                TreeName = treeName,
+                Dependencies = new List<Type>() { typeof(FooEnum) },
+                ScriptCache = this.scriptCache
+            };
+
+            this.session = new TreeWalkerSession(this.parameters);
+        }
+
+        /// <summary>
+        /// Helper function to test V1 ITreeWalkerCallbacks on the contructor accepting FileSchema.
```
TravisOnGit, line 73: >FileSchema [](start = 88, length = 10)

jsonSchema

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
TravisOnGit: The "VisitNode" tests are at the top, followed by "WalkTree" tests.

Please remove this VisitNode test and add a new test just under TestTreeWalkerSession_WalkTree_Success.

---
In reply to: [505946715](https://github.com/microsoft/Forge/pull/32#discussion_r505946715) [](ancestors = 505946715,505916607)

Forge.TreeWalker/src/ITreeWalkerCallbacks.cs
```diff
@@ -14,7 +14,9 @@ namespace Microsoft.Forge.TreeWalker
     using System.Threading.Tasks;
 
     /// <summary>
+    /// [DEPRECATED]
     /// The ITreeWalkerCallbacks interface defines the callback Tasks that are awaited while walking the tree.
+    /// Prefers future callers to use the new ITreeWalkerCallbacksV2, and keep this ITreeWalkerCallbacks for backward compatibility.
```
TravisOnGit, line 19: >, and keep this ITreeWalkerCallbacks for backward compatibility [](start = 68, length = 63)

Remove this comment.

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,102 @@
...
+// </copyright>
+// <summary>
+//     The TreeNodeContext class.
+// </summary>
+//-----------------------------------------------------------------------
+
+namespace Microsoft.Forge.TreeWalker
+{
+    using System;
+    using System.Threading;
```
TravisOnGit, line 13: >using System.Threading; [](start = 4, length = 23)

I don't think this using statement is required either, please remove if possible.

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// </summary>
+        /// <param name="sessionId">The unique identifier for this tree walking session.</param>
+        /// <param name="treeNodeKey">The tree node's key.</param>
+        /// <param name="properties">The properties of this tree node.</param>
+        /// <param name="userContext">Optional, the user context for this tree node.</param>
+        /// <param name="token">The cancellation token.</param>
+        /// <param name="treeName">The name of the ForgeTree in the JsonSchema.</param>
+        /// <param name="rootSessionId">The unique identifier for the root/parent tree walking session.</param>
+        /// <param name="currentNodeSkipActionContext">
+        /// The string context if the actions in this tree node are skipped. It is null when no skip. When it is not empty, proceed with ChildSelector.
```
TravisOnGit, line 74: >The string context if the actions in this tree node are skipped. It is null when no skip. When it is not empty, proceed with ChildSelector. [](start = 12, length = 139)

The string context if the actions in the current tree node should be skipped, or null if actions should not be skipped.

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <summary>
+        /// Instantiates an TreeNodeContext object.
+        /// </summary>
+        /// <param name="sessionId">The unique identifier for this tree walking session.</param>
+        /// <param name="treeNodeKey">The tree node's key.</param>
+        /// <param name="properties">The properties of this tree node.</param>
+        /// <param name="userContext">Optional, the user context for this tree node.</param>
+        /// <param name="token">The cancellation token.</param>
+        /// <param name="treeName">The name of the ForgeTree in the JsonSchema.</param>
+        /// <param name="rootSessionId">The unique identifier for the root/parent tree walking session.</param>
```
TravisOnGit, line 72: Let's re-use the same param summary from ITreeWalkerCallbacks.

Forge.TreeWalker/src/ITreeWalkerCallbacksV2.cs
```diff
@@ -0,0 +1,31 @@
...
+
+    /// <summary>
+    /// The ITreeWalkerCallbacks interface defines the callback Tasks that are awaited while walking the tree.
+    /// </summary>
+    public interface ITreeWalkerCallbacksV2
+    {
+        /// <summary>
+        /// The callback Task that is awaited before visiting each node.
+        /// </summary>
+        /// <param name="treeNodeContext">The TreeNode context holding relevant information for this Action.</param>
```
TravisOnGit, line 22: >The TreeNode context holding relevant information for this Action. [](start = 42, length = 66)

The tree node context holding relevant information about this tree node and session.

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,102 @@
...
+// </summary>
+//-----------------------------------------------------------------------
+
+namespace Microsoft.Forge.TreeWalker
+{
+    using System;
+    using System.Threading;
+
+    /// <summary>
+    /// The TreeNodeContext class object is passed to Callbacks.BeforeVisitNode and Callbacks.AfterVisitNode.
```
TravisOnGit, line 16: >The TreeNodeContext class object is passed to Callbacks.BeforeVisitNode and Callbacks.AfterVisitNode. [](start = 8, length = 101)

The TreeNodeContext class holds relevant information about the tree node and session.
This object gets passed to CallbacksV2.BeforeVisitNode and CallbacksV2.AfterVisitNode.

Forge.TreeWalker/src/TreeNodeContext.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        /// <summary>
+        /// The name of the ForgeTree in the JsonSchema.
+        /// </summary>
+        public string TreeName { get; private set; }
+
+        /// <summary>
+        /// The unique identifier for the root/parent tree walking session.
+        /// </summary>
+        public Guid RootSessionId { get; private set; }
```
TravisOnGit, line 53: Let's use the summaries from ITreeWalkerCallbacks.

Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -180,8 +192,48 @@ public class TreeWalkerParameters
             this.SessionId = sessionId;
             this.JsonSchema = jsonSchema;
             this.ForgeState = forgeState;
+            this.Token = token;
             this.Callbacks = callbacks;
```
TravisOnGit, line 196: Flip the order here. Callbacks before token to line up with parameters.

Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -151,11 +162,12 @@ public class TreeWalkerParameters
             this.SessionId = sessionId;
             this.ForgeTree = forgeTree;
             this.ForgeState = forgeState;
-            this.Callbacks = callbacks;
             this.Token = token;
+            this.Callbacks = callbacks;
```
TravisOnGit, line 166: Flip the order here. Callbacks before token to line up with parameters.

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: fixed. (i saw some duplication and unclear in the iteration #7)

---
In reply to: [506620610](https://github.com/microsoft/Forge/pull/32#discussion_r506620610) [](ancestors = 506620610,505924425,505910307,505857375,505841676,505811771,505150928,505093484,504257524,504223272)

Forge.TreeWalker/src/ITreeWalkerCallbacks.cs
zhengmu: removed

---
In reply to: [506628449](https://github.com/microsoft/Forge/pull/32#discussion_r506628449) [](ancestors = 506628449)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: CancellationToken is from "System.Threading". Need to keep as it is.

---
In reply to: [506629130](https://github.com/microsoft/Forge/pull/32#discussion_r506629130) [](ancestors = 506629130)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: updated

---
In reply to: [506634857](https://github.com/microsoft/Forge/pull/32#discussion_r506634857) [](ancestors = 506634857)

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: updated

---
In reply to: [506636201](https://github.com/microsoft/Forge/pull/32#discussion_r506636201) [](ancestors = 506636201)

Forge.TreeWalker/src/TreeWalkerParameters.cs
zhengmu: updated

---
In reply to: [506636335](https://github.com/microsoft/Forge/pull/32#discussion_r506636335) [](ancestors = 506636335)

Forge.TreeWalker/src/ITreeWalkerCallbacksV2.cs
zhengmu: updated. applied the same to AfterVisitNode() below.

---
In reply to: [506633634](https://github.com/microsoft/Forge/pull/32#discussion_r506633634) [](ancestors = 506633634)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: updated

---
In reply to: [506631283](https://github.com/microsoft/Forge/pull/32#discussion_r506631283) [](ancestors = 506631283)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: fixed

---
In reply to: [506632409](https://github.com/microsoft/Forge/pull/32#discussion_r506632409) [](ancestors = 506632409)

Forge.TreeWalker/src/TreeNodeContext.cs
zhengmu: fixed

---
In reply to: [506635372](https://github.com/microsoft/Forge/pull/32#discussion_r506635372) [](ancestors = 506635372)

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
zhengmu: fixed.

---
In reply to: [506626078](https://github.com/microsoft/Forge/pull/32#discussion_r506626078) [](ancestors = 506626078)

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
zhengmu: Just noticed the name ordering, I renamed the two cases and moved them into *WalkTree" session

---
In reply to: [506627465](https://github.com/microsoft/Forge/pull/32#discussion_r506627465) [](ancestors = 506627465,505946715,505916607)

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
zhengmu: moved the V1 initializers to follow the regular initializers.

---
In reply to: [506625855](https://github.com/microsoft/Forge/pull/32#discussion_r506625855) [](ancestors = 506625855)

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
zhengmu: updated

---
In reply to: [506625115](https://github.com/microsoft/Forge/pull/32#discussion_r506625115) [](ancestors = 506625115)

TravisOnGit (approved):
:shipit:

TravisOnGit (approved):
:shipit:

PR #41, opened 2021-03-02
NetStandard workaround for MissingResolver issue plus other cleanups

TravisOnGit:
There's a known issue with NetStandard/NetCore when overriding the MetadataReferenceResolver. I worked around this by using conditional compilation to only use the MissingResolver when the target framework is Net462.

Added PackageTags to csproj for better discoverability in NuGet.

Fixing up some random warnings.


Forge.TreeWalker/src/ExpressionExecutor.cs
```diff
@@ -146,13 +146,18 @@ private void Initialize()
...
+                this.ParentScriptTask = Task.CompletedTask;
                 return;
             }
 
-            this.parentScriptTask = Task.Run(async () =>
+            this.ParentScriptTask = Task.Run(async () =>
             {
-                ScriptOptions scriptOptions = ScriptOptions.Default.WithMetadataResolver(new MissingResolver());
+                ScriptOptions scriptOptions = ScriptOptions.Default;
+#if NET462
```
TravisOnGit, line 156: Only actual change is here.

zhengmu (approved):
:shipit:

PR #46, opened 2021-08-31
Added RetryCurrentTreeNodeActions flag support to re-execute node upo…

kijujjav:
Added RetryCurrentTreeNodeActions flag support to re-execute node upon rehydration


Forge.TreeWalker/src/TreeWalkerParameters.cs
```diff
@@ -132,6 +132,11 @@ public class TreeWalkerParameters
         /// </summary>
         public object TreeInput { get; set; }
 
+        /// <summary>
+        /// Flag that will allow forge to re-execute the Node upon re-hydration.
```
TravisOnGit, line 136: >Flag that will allow forge to re-execute the Node upon re-hydration [](http://example.com/codeflow?start=12&length=67)

Flag that allows Forge to re-execute the current TreeNode's Actions upon rehydration.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -1123,7 +1123,7 @@ private async Task CommitCurrentTreeNode(string treeNodeKey, TreeNode treeNode)
             List<KeyValuePair<string, object>> itemsToPersist = new List<KeyValuePair<string, object>>();
 
             // Handle revisit/cycle case if this node has been previously completed and we aren't rehydrating.
```
TravisOnGit, line 1125: >and we aren't rehydrating [](http://example.com/codeflow?start=84&length=25)

"and we are rehydrated or the override flag is set."

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -500,12 +500,71 @@ public void TestTreeWalkerSession_WalkTree_Failed_MissingKey()
...
         {
             this.TestInitialize(jsonSchema: ForgeSchemaHelper.NoChildMatch);
-
             // Test - WalkTree and expect the Status to be NoChildMatched.
             string actualStatus = this.session.WalkTree("Root").GetAwaiter().GetResult();
             Assert.AreEqual("RanToCompletion_NoChildMatched", actualStatus, "Expected WalkTree to end with NoChildMatched status.");
         }
 
+        [TestMethod]
+        public void TestReexecutingNode_WithRetryCurrentTreeNodeActionsFlag_Success()
```
TravisOnGit, line 509: >TestReexecutingNode_WithRetryCurrentTreeNodeActionsFlag_Success [](http://example.com/codeflow?start=20&length=63)

nit: You flipped the names of your tests. The first test is actually without the flag.

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -500,12 +500,71 @@ public void TestTreeWalkerSession_WalkTree_Failed_MissingKey()
...
+                parameters.ForgeState,
+                this.callbacksV2,
+                this.token)
+            {
+                UserContext = new ForgeUserContext(),
+                ForgeActionsAssembly = typeof(CollectDiagnosticsAction).Assembly,
+                InitializeSubroutineTree = this.InitializeSubroutineTree,
+                Dependencies = new List<Type>() { typeof(FooEnum) },
+                ScriptCache = this.scriptCache
+            });
```
TravisOnGit, line 528: I think you can just do this:

this.session = new TreeWalkerSession(this.parameters);

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -500,12 +500,71 @@ public void TestTreeWalkerSession_WalkTree_Failed_MissingKey()
...
             string actualStatus = this.session.WalkTree("Root").GetAwaiter().GetResult();
             Assert.AreEqual("RanToCompletion_NoChildMatched", actualStatus, "Expected WalkTree to end with NoChildMatched status.");
         }
 
+        [TestMethod]
+        public void TestReexecutingNode_WithRetryCurrentTreeNodeActionsFlag_Success()
+        {
+            this.TestInitialize(jsonSchema: ForgeSchemaHelper.ReExecuteNodeSchema);
+
+            this.session.WalkTree("Root").GetAwaiter().GetResult();
```
TravisOnGit, line 513: >this.session.WalkTree("Root").GetAwaiter().GetResult(); [](http://example.com/codeflow?start=12&length=55)

Let's add a comment for clarity and consistency:

// Test - WalkTree twice without the RetryCurrentTreeNodeActions flag set. Expect the Action to not get re-executed since the ActionResponse already exists and we are rehydrating.

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -500,12 +500,71 @@ public void TestTreeWalkerSession_WalkTree_Failed_MissingKey()
...
+                this.callbacksV2,
+                this.token)
+            {
+                UserContext = new ForgeUserContext(),
+                ForgeActionsAssembly = typeof(CollectDiagnosticsAction).Assembly,
+                InitializeSubroutineTree = this.InitializeSubroutineTree,
+                Dependencies = new List<Type>() { typeof(FooEnum) },
+                ScriptCache = this.scriptCache,
+                RetryCurrentTreeNodeActions = true
+            });
```
TravisOnGit, line 559: I think we can do this:

this.parameters.RetryCurrentTreeNodeActions = true;
this.session = new TreeWalkerSession(this.parameters);

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -500,12 +500,71 @@ public void TestTreeWalkerSession_WalkTree_Failed_MissingKey()
...
+
+            this.session.WalkTree("Root").GetAwaiter().GetResult();
+            string time2 = session.GetLastActionResponse().Status;
+
+            Assert.IsFalse(time1 == time2);
```
TravisOnGit, line 564: >Assert.IsFalse(time1 == time2); [](http://example.com/codeflow?start=12&length=31)

Instead of adding a new test Action, could we re-use the RevisitAction? I'm always a bit cautious about relying on DateTime.UtcNow in UTs because they can execute so quickly and create weirdness.

RevisitAction works like this. Every time it is executed, it looks at the PreviousActionResponse and increases it's ActionResponse.Output by 1.

In the first UT (without the new flag set), we would expect both outputs to be 0 (because the Action doesn't get re-executed).

In the second UT (with the new flag set), we would expect the first output to be 0 and the retry to be 1 (because the Action was re-executed).

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -500,12 +500,71 @@ public void TestTreeWalkerSession_WalkTree_Failed_MissingKey()
...
+                Dependencies = new List<Type>() { typeof(FooEnum) },
+                ScriptCache = this.scriptCache
+            });
+
+            this.session.WalkTree("Root").GetAwaiter().GetResult();
+            string time2 = session.GetLastActionResponse().Status;
+
+            Assert.IsTrue(time1 == time2);
+
+        }
```
TravisOnGit, line 535: nit: remove newline before closing brace. Same for other UT.

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -500,12 +500,49 @@ public void TestTreeWalkerSession_WalkTree_Failed_MissingKey()
         public void TestTreeWalkerSession_WalkTree_NoChildMatched()
         {
             this.TestInitialize(jsonSchema: ForgeSchemaHelper.NoChildMatch);
-
             // Test - WalkTree and expect the Status to be NoChildMatched.
             string actualStatus = this.session.WalkTree("Root").GetAwaiter().GetResult();
```
TravisOnGit, line 504: nit: Can you add this newline back above this comment?

Forge.TreeWalker.UnitTests/test/TreeWalkerUnitTests.cs
```diff
@@ -500,12 +500,49 @@ public void TestTreeWalkerSession_WalkTree_Failed_MissingKey()
...
             Assert.AreEqual("RanToCompletion_NoChildMatched", actualStatus, "Expected WalkTree to end with NoChildMatched status.");
         }
 
+        [TestMethod]
+        // WalkTree twice without the RetryCurrentTreeNodeActions flag set.
+        // Expect the Action to not get re-executed since the ActionResponse already exists and we are rehydrating.
```
TravisOnGit, line 510: nit: Please place this comment above this.session.WalkTree. That's the style in this UT file. Also please preface the comment with "Test - "

TravisOnGit (approved):
:shipit:

PR #55, opened 2023-08-22
Implemented Try/Catch for Improved Clarity in ShouldSelect Expression…

alekssaf:
… Errors

### DIFF
--- a/Forge.TreeWalker/src/TreeWalkerSession.cs
+++ b/Forge.TreeWalker/src/TreeWalkerSession.cs
@@ -534,7 +534,7 @@ namespace Microsoft.Forge.TreeWalker
                 {
                     return cs.Child;
                 }
-                var evaluationResult = await EvaluateDynamicProperty(cs.ShouldSelect, typeof(bool)).ConfigureAwait(false);
+                var evaluationResult = await this.EvaluateDynamicProperty(cs.ShouldSelect, typeof(bool)).ConfigureAwait(false);
                 try
                 {
                     if ((bool)evaluationResult)
@@ -544,7 +544,7 @@ namespace Microsoft.Forge.TreeWalker
                 }
                 catch (Exception e)
                 {
-                    throw new Exception($"An error occurred while casting 'ShouldSelect' expression: '{cs.ShouldSelect}' in the ChildSelector with child '{cs.Child}': '{e.Message}'.", e);
+                    throw new Exception($"An error occurred while casting 'ShouldSelect' expression: '{cs.ShouldSelect}' in the ChildSelector with child '{cs.Child}'. Exception: '{e.Message}'.", e);
                 }
             }
 
--- a/Forge.TreeWalker/src/TreeWalkerSession.cs
+++ b/Forge.TreeWalker/src/TreeWalkerSession.cs
@@ -534,9 +534,10 @@ namespace Microsoft.Forge.TreeWalker
                 {
                     return cs.Child;
                 }
+                var evaluationResult = await EvaluateDynamicProperty(cs.ShouldSelect, typeof(bool)).ConfigureAwait(false);
                 try
                 {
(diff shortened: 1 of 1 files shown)


alekssaf:
https://github.com/microsoft/Forge/pull/55

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -534,9 +534,16 @@ internal async Task<string> SelectChild(TreeNode treeNode)
...
                 {
-                    return cs.Child;
+                    if ((bool)await EvaluateDynamicProperty(cs.ShouldSelect, typeof(bool)).ConfigureAwait(false))
+                    {
+                        return cs.Child;
+                    }
+                }
+                catch (Exception e)
+                {
+                    throw new Exception($"An error occurred while evaluating 'ShouldSelect' expression: '{cs.ShouldSelect}' in the ChildSelector with child '{cs.Child}': '{e.Message}'.", e);
```
TravisOnGit: Hmm.. My one concern is that this change would change the behavior when we DO hit an EvaluateDynamicPropertyException when calling EvaluateDynamicProperty. We would throw an Exception instead, which leads to a "Failed" Status instead of "Failed_EvaluateDynamicProperty". Perhaps if we broke the statement into 2 lines. One line to call EvaluateDynamicProperty. Then inside a try/catch, cast to bool and return.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -534,9 +534,16 @@ internal async Task<string> SelectChild(TreeNode treeNode)
                 {
                     return cs.Child;
                 }
-                if ((bool)await this.EvaluateDynamicProperty(cs.ShouldSelect, typeof(bool)).ConfigureAwait(false))
+                try
                 {
-                    return cs.Child;
+                    if ((bool)await EvaluateDynamicProperty(cs.ShouldSelect, typeof(bool)).ConfigureAwait(false))
```
TravisOnGit: nit: Be sure to add back "this."

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -534,9 +534,17 @@ internal async Task<string> SelectChild(TreeNode treeNode)
...
                 {
-                    return cs.Child;
+                    if ((bool)evaluationResult)
+                    {
+                        return cs.Child;
+                    }
+                }
+                catch (Exception e)
+                {
+                    throw new Exception($"An error occurred while casting 'ShouldSelect' expression: '{cs.ShouldSelect}' in the ChildSelector with child '{cs.Child}': '{e.Message}'.", e);
```
TravisOnGit, line 547: $"An error occurred while casting ShouldSelect expression: '{cs.ShouldSelect}' in the ChildSelector with child: '{cs.Child}'. Exception: {e.Message}."

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -534,9 +534,17 @@ internal async Task<string> SelectChild(TreeNode treeNode)
                 {
                     return cs.Child;
                 }
-                if ((bool)await this.EvaluateDynamicProperty(cs.ShouldSelect, typeof(bool)).ConfigureAwait(false))
+                var evaluationResult = await EvaluateDynamicProperty(cs.ShouldSelect, typeof(bool)).ConfigureAwait(false);
+                try
```
TravisOnGit, line 538: Your branch name should have the name: feature/foo
That way, the CI/CD pipelines will automatically pick up your branch and PR. Please try sending a new PR with correct branch name format, ty!

TravisOnGit (commented):
Left some comments, please take a look.

TravisOnGit (approved):

PR #70, opened 2025-08-26
Fix #69: Throw exception for missing actions referenced in schema

### LINKED ISSUE #69: No Exception Thrown for Missing Action in Schema – GetLastActionResponse() Uses Previous Action's Response
We noticed a potentially risky behavior while working with Forge.

When using an Action defined in the schema that does not exist in the code implementation, Forge doesn't throw any error or warning. Instead, the flow continues as if the action executed successfully.

Even more concerning, when GetLastActionResponse() is called afterward, it seems to return the response from the previous valid action, not indicating that the last action was actually skipped or failed. This behavior can silently introduce logic bugs, especially when ShouldSelect statements rely on the response from a specific action.

Ideally, Forge should:
1. Throw a clear error or warning if an action is referenced in the schema but not implemented.
2. Ensure GetLastActionResponse() reflects the actual last attempted action, or explicitly indicates that the last action failed or was invalid.

SatishUpadhyaya:
CLOSES: #69

### DIFF
--- a/Forge.TreeWalker/src/TreeWalkerSession.cs
+++ b/Forge.TreeWalker/src/TreeWalkerSession.cs
@@ -13,6 +13,7 @@ namespace Microsoft.Forge.TreeWalker
     using System.Collections;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.Linq;
     using System.Reflection;
     using System.Text.RegularExpressions;
     using System.Threading;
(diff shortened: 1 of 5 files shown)


Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -13,6 +13,7 @@ namespace Microsoft.Forge.TreeWalker
     using System.Collections;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.Linq;
```
TravisOnGit, line 16: Please remove unused reference to System.Linq, ty!

SatishUpadhyaya:
@microsoft-github-policy-service agree company="Microsoft"

SatishUpadhyaya:
All tests passed locally:
{76D65861-B6B0-482B-B6D4-42002122E512}

kijujjav (approved):

TravisOnGit (approved):

PR #86, opened 2026-05-18
Update GitVersionTask package version to 5.2.4

lakshmipriya-gottipati:


TravisOnGit (approved):
Approved! There may be more nuance needed later but for now we have a passing build again. Thanks for unblocking the pipeline failures!

PR #87, opened 2026-06-05
Add CacheVariables feature to TreeWalker

ilang898:
## Summary

Adds a new **CacheVariables** property to TreeNode that enables declarative, node-scoped variable caching within workflow schemas. Schema authors can evaluate expressions after actions complete and bind results to named variables accessible in ShouldSelect routing expressions via the Cache object.

## Problem

Current ShouldSelect expressions must reference action outputs inline with repeated Session.GetOutput() calls, leading to verbose, hard-to-read routing logic.

## Solution

\\\json
"CacheVariables": {
    "scsSucceeded": "C#|Session.GetOutput(\"GetBypassSafetyCheckToken\").Output.Succeeded",
    "scsIsSafe": "C#|Session.GetOutput(\"GetBypassSafetyCheckToken\").Output.IsSafe"
},
"ChildSelector": [
    { "ShouldSelect": "C#|Cache.scsSucceeded && Cache.scsIsSafe", "Child": "Next" }
]
\\\

## Key Design Decisions

- **Node-scoped**: Cache is cleared at the start of each \VisitNode\ — no cross-node leakage, no persistence to ForgeState
- **Uses existing EvaluateDynamicProperty**: Typed as \dynamic\, evaluated the same way as \Properties\, \Input\, and \TreeInput\
- **JObject-backed Cache**: Supports \Cache.x\ dot-notation in Roslyn expressions via \IDynamicMetaObjectProvider\
- **Fully backward compatible**: Optional property (null by default), no changes to existing contracts or behavior

## Files Changed

| File | Change |
|------|--------|
| \ForgeTree.cs\ | Added \CacheVariables\ (dynamic) to \TreeNode\ |
| \ExpressionExecutor.cs\ | Added \Cache\ property, \SetCache\/\ClearCache\/\GetCache\ methods |
| \ITreeSession.cs\ | Added \GetCache(string name)\ to interface |
| \TreeWalkerSession.cs\ | Cache clear/evaluate/set in \VisitNode\, \GetCache\ implementation |
| \ForgeSchemaValidationRules.json\ | Added \CacheVariables\ to node definitions |

## Testing

- 10 new unit tests covering: static expressions, Session.GetOutput access, UserContext binding, node-scoped isolation, invalid expressions, multiple variables, string literals, GetCache API, backward compatibility
- All 105 tests pass (95 original + 10 new)


Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -477,6 +505,9 @@ public async Task<string> VisitNode(string treeNodeKey)
         {
             TreeNode treeNode = this.Schema.Tree[treeNodeKey];
 
+            // Clear node-scoped cache variables at the start of each node visit.
+            this.expressionExecutor.ClearCache();
```
TravisOnGit: Let's consider adding this ClearCache call directly after awaiting SelectChild. Doing that would definitely lock it down to only being used during SelectChild.

Having ClearCache at the top of VisitNode here means that AfterVisitNode would also have access to the Cache. And also if folks directly call VisitNode (without WalkTree), the Cache would be maintained after returning VisitNode, so app (or UTs) could then access the Cache as well.

Let's discuss offline if there are valid use-cases or any drawbacks to calling ClearCache at the beginning of VisitNode vs at the end.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -508,6 +539,11 @@ public async Task<string> VisitNode(string treeNodeKey)
                 }
             }
 
+            // Evaluate CacheVariables after actions complete, before selecting child.
+            // Uses the same EvaluateDynamicProperty pattern as Properties/Input.
+            object cacheResult = await this.EvaluateDynamicProperty(treeNode.CacheVariables, null).ConfigureAwait(false);
```
TravisOnGit: nit: Let's add this after the if Leaf statement. Since it won't be needed if that returns.

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -508,6 +539,11 @@ public async Task<string> VisitNode(string treeNodeKey)
...
+            this.expressionExecutor.SetCache(cacheResult);
```
TravisOnGit: Idea to have a SetCache method in TreeWalkerSession. It would do these 2 operations, EvaluateDynamicProperty and SetCache in ExpressionExecutor.

The benefit of having a SetCache method is for UTs that validate schemas by evaluating all the expressions. For example, a UT that evaluates the ShouldSelect statements by calling EvaluateDynamicProperty. With the new Cache feature, ShouldSelect statements would fail unless the Cache is filled with the expected values. So we need some way to update the Cache with expected values. Something like:
`await treeWalkerSession.SetCache(forgeTreeNode.CacheVariables);
await treewalkerSession.EvaluateDynamicProperty(childSelector, typeof(bool));`

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -282,6 +282,34 @@ public string GetCurrentNodeSkipActionContext()
             return this.currentNodeSkipActionContext;
         }
 
+        /// <summary>
+        /// Gets the value of a node-scoped cache variable by name.
+        /// Cache variables are populated from CacheVariables expressions after actions complete.
+        /// </summary>
+        /// <param name="name">The variable name as defined in CacheVariables.</param>
+        /// <returns>The cached value if it exists, otherwise null.</returns>
+        public object GetCache(string name)
```
TravisOnGit: Is this called anywhere? I don't understand the need for this method.

Forge.TreeWalker/src/ExpressionExecutor.cs
```diff
@@ -79,7 +79,8 @@ public ExpressionExecutor(ITreeSession session, object userContext, List<Type> d
             {
                 UserContext = userContext,
                 Session = session,
-                TreeInput = treeInput
+                TreeInput = treeInput,
+                Cache = new Newtonsoft.Json.Linq.JObject()
```
TravisOnGit: TreeInput is already dynamic and is handled. Can we have Cache treated similarly? I don't think we need to set it as a JObject or add an additional assembly for that.

I think setting to null here is fine. If folks accidentally try to use the Cache (from ShouldSelect) when there's nothing inside (Cache in the schema), it would throw a NullReferenceException, which should be pretty clear that it's because there's no Cache defined (in the schema). Versus if we set this Cache to a default Object instead, it would throw a more confusing exception in the same scenario.

Forge.TreeWalker/src/ExpressionExecutor.cs
```diff
@@ -208,6 +210,32 @@ public bool ScriptCacheContainsKey(string expression)
...
+        {
+            this.parameters.Cache = cache;
+        }
+
+        /// <summary>
+        /// Clears the Cache, resetting it to an empty JObject for the next node visit.
+        /// </summary>
+        public void ClearCache()
+        {
+            this.parameters.Cache = new Newtonsoft.Json.Linq.JObject();
```
TravisOnGit: I think setting to null here is fine.

Forge.TreeWalker/src/ExpressionExecutor.cs
```diff
@@ -208,6 +210,32 @@ public bool ScriptCacheContainsKey(string expression)
             return this.scriptCache.ContainsKey(expression);
         }
 
+        /// <summary>
+        /// Gets the Cache object (a JObject with evaluated CacheVariable values, or null).
+        /// </summary>
+        /// <returns>The cache object.</returns>
+        public dynamic GetCache()
```
TravisOnGit: I _think_ this can return "object" type instead of dynamic. I see that GetOrCommitTreeInput returns object.

Forge.TreeWalker/contracts/ForgeTree.cs
```diff
@@ -58,6 +58,16 @@ public class TreeNode
...
 
+        /// <summary>
+        /// Optional cache variables that are evaluated via EvaluateDynamicProperty after actions complete
+        /// and made available in ShouldSelect expressions via the Cache object (e.g., "Cache.myVar").
+        /// Each property key is the variable name, and the value is an expression
+        /// (e.g., "C#|Session.GetOutput(\"ActionKey\").Output.PropertyName").
+        /// Cache variables are scoped to the current node only — they are cleared when moving to the next node.
+        /// </summary>
+        [DataMember]
+        public dynamic CacheVariables { get; private set; }
```
TravisOnGit: Consider just "Cache" for the name.

Forge.TreeWalker/contracts/ForgeSchemaValidationRules.json
```diff
@@ -50,6 +50,12 @@
                 },
                 "Properties": {
                     "type": "object"
+                },
+                "CacheVariables": {
```
TravisOnGit: I think this is right, setting type as object and setting patternProperties. This should effectively enforce it is used like a dictionary with named, string properties. As opposed to "Properties" above that's simply an object. It can be set to a dynamic object, a string, an int, a dictionary, etc..

In theory, we could also set CacheVariables to object type like Properties. It would be more flexible/versatile for users, like if they want to simply set a single value instead of using the named properties. But I think I like the guardrails of enforcing a dictionary-style pattern. It'll make it more consistent and less error-prone.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -570,5 +570,319 @@ public static class ForgeSchemaHelper
...
+
+        /// <summary>
+        /// Basic CacheVariables test — static Roslyn expression evaluated after action, used in ShouldSelect.
+        /// </summary>
+        public const string CacheVariables_StaticExpression = @"
+            {
+                ""Tree"": {
+                    ""Root"": {
+                        ""Type"": ""Action"",
+                        ""Actions"": {
```
TravisOnGit: Recommend removing the Actions and just making these Selection TreeNode types, unless you are utilizing the ActionResponse. Only include things that contribute to each UT in some way and cut the fluff please.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -570,5 +570,319 @@ public static class ForgeSchemaHelper
...
+                        ""ChildSelector"": [
+                            {
+                                ""Child"": ""SecondNode""
+                            }
+                        ]
+                    },
+                    ""SecondNode"": {
+                        ""Type"": ""Selection"",
+                        ""CacheVariables"": {
+                            ""secondNodeCheck"": ""C#|Session.GetCache(\""firstNodeVar\"") == null ? \""isolated\"" : \""leaked\""""
```
TravisOnGit: Session.GetCache I don't believe is a use-case. You should not be able to access the Cache from inside the Cache. Please remove this.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -570,5 +570,319 @@ public static class ForgeSchemaHelper
...
+                    ""NotFound"": {
+                        ""Type"": ""Leaf""
+                    }
+                }
+            }";
+
+        /// <summary>
+        /// CacheVariables with string literal (non-Roslyn value).
+        /// </summary>
+        public const string CacheVariables_StringLiteral = @"
```
TravisOnGit: More UT ideas:

- CacheVariables_IsString_ExpectException - Set the Cache directly to a string instead of a dictionary, like: ""Cache"": ""value"". Maybe put it in a separate InvalidTestSchemas and confirm exception is thrown. Or maybe a forge schema validation test would be appropriate to ensure schema validation catches that issue.
- Set a CacheVariable value as an object, like ActionResponse. Then confirm ShouldSelect can use it like Cache.DiagnosticsActionResponse.Status
- Set a CacheVariable value as a boolean expression like Session.GetOutput(\""Root_CollectDiagnosticsAction\"").Status == \""Success\"". Then confirm ShouldSelect can use it like Cache.IsSuccess.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -570,5 +570,319 @@ public static class ForgeSchemaHelper
...
+                                ""Action"": ""CollectDiagnosticsAction"",
+                                ""Input"": {
+                                    ""Command"": ""TheCommand""
+                                }
+                            }
+                        },
+                        ""CacheVariables"": {
+                            ""a"": ""C#|10"",
+                            ""b"": ""C#|20"",
+                            ""sum"": ""C#|30""
```
TravisOnGit: nit: It's minor, but it could lead to confusion. You list variables, "a", "b", and "sum". This could imply that you could calculate sum by USING what you just set in a and b. But this is not possible, you can't view what's in the Cache from CacheVariables. So just to avoid that possible confusion, can you change the example?

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
TravisOnGit: - Using all the TreeNode Types, like Selection, Leaf, and Subroutine. Just simple tests to confirm Cache works on all the types.

Forge.TreeWalker/contracts/ForgeTree.cs
TravisOnGit: There's an argument that "CacheVariables" is more clear to what it does. At first glance at a schema, who's to say what a "Cache" could represent..
"Variables" takes forever to say though, 😄.. Since this is what is shown to the user in the schema, let's brainstorm best name. Here's some ideas:

1. Cache
2. CacheVariables
3. CacheVars
4. TreeNodeCache

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
TravisOnGit: - Have 2 TreeNodes that have the same exact named property in their Cache. Confirm that the second TreeNode's ShouldSelect statement contains its expected value. This is a good confirmation that the Cache truly gets wiped.

TravisOnGit (commented):
Left feedback on iteration 7. Please have a look, thanks!

Forge.TreeWalker/contracts/ForgeTree.cs
ilang898: I think CacheVars is a good compromise between being descriptive and concise.

Forge.TreeWalker/src/TreeWalkerSession.cs
ilang898: I agree this feature is intended to be scoped to the ShouldSelectors, so we should move the ClearCache to the end so that it cannot be used after the selectors have been processed.

Forge.TreeWalker/src/TreeWalkerSession.cs
ilang898: Done

Forge.TreeWalker/src/TreeWalkerSession.cs
ilang898: Added the suggested SetCache method

Forge.TreeWalker/src/TreeWalkerSession.cs
ilang898: This was leftover from my initial implementation. Now removed

Forge.TreeWalker/src/ExpressionExecutor.cs
ilang898: Updated to initialize as null

Forge.TreeWalker/src/ExpressionExecutor.cs
ilang898: Setting to null instead of new object

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -714,8 +690,8 @@ public static class ForgeSchemaHelper
                     },
                     ""SecondNode"": {
                         ""Type"": ""Selection"",
-                        ""CacheVariables"": {
-                            ""secondNodeCheck"": ""C#|Session.GetCache(\""firstNodeVar\"") == null ? \""isolated\"" : \""leaked\""""
+                        ""CacheVars"": {
+                            ""secondNodeCheck"": ""C#|Cache == null ? \""isolated\"" : \""leaked\""""
```
TravisOnGit: This is an invalid use-case, right? We shouldn't be calling C#|Cache from inside CacheVars.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -846,43 +836,133 @@ public static class ForgeSchemaHelper
...
+                        ""Type"": ""Leaf""
+                    },
+                    ""NotReady"": {
+                        ""Type"": ""Leaf""
+                    }
+                }
+            }";
+
+        /// <summary>
+        /// CacheVars on multiple node types — Selection and Action.
```
TravisOnGit: Works for Subroutine too, right? Add UT for that as well please.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -823,12 +813,12 @@ public static class ForgeSchemaHelper
...
                         },
-                        ""CacheVariables"": {
-                            ""literal"": ""hello world""
+                        ""CacheVars"": {
+                            ""response"": ""C#|Session.GetOutput(\""Root_CollectDiagnosticsAction\"")""
                         },
                         ""ChildSelector"": [
                             {
-                                ""ShouldSelect"": ""C#|Cache.literal.ToString() == \""hello world\"""",
+                                ""ShouldSelect"": ""C#|Cache.response != null"",
```
TravisOnGit: Can we take this further? What's it look like if the Output is a custom object (Like the ActionResponse.Output set to a custom type)? Can we use it directly, or maybe we need some casting? It's handy to have use cases in the UTs to refer to when folks ask what is possible.
Cache.Output.Choice == "ABC"
((ChoiceObject)Cache.Output).Choice == "ABC"

TravisOnGit: Also in this current test, have the ShouldSelect statement check if Cache.response.Status == "Success". Just an extra test of using that known object.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -846,43 +836,133 @@ public static class ForgeSchemaHelper
...
+        /// CacheVars with boolean expression — use Cache.IsReady in ShouldSelect.
         /// </summary>
-        public const string CacheVariables_GetCacheApi = @"
+        public const string CacheVars_BooleanExpression = @"
             {
                 ""Tree"": {
                     ""Root"": {
+                        ""Type"": ""Selection"",
+                        ""CacheVars"": {
+                            ""IsReady"": ""C#|true""
```
TravisOnGit: Instead of setting to true like this, have it evaluate an expression. Like
"IsSuccess": "C#|Session.GetOutput(\""Root_CollectDiagnosticsAction\"").Status == "Success""

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
ilang898: Yeah I'll update this to perform the null check on the variable itself in the child selector

Forge.TreeWalker/src/TreeWalkerSession.cs
```diff
@@ -283,31 +283,14 @@ public string GetCurrentNodeSkipActionContext()
...
-        /// Gets the value of a node-scoped cache variable by name.
-        /// Cache variables are populated from CacheVariables expressions after actions complete.
+        /// Evaluates the given CacheVars schema object and sets the result on the Cache.
+        /// This is useful for external UTs that need to pre-fill the Cache before evaluating ShouldSelect expressions.
         /// </summary>
-        /// <param name="name">The variable name as defined in CacheVariables.</param>
-        /// <returns>The cached value if it exists, otherwise null.</returns>
-        public object GetCache(string name)
+        /// <param name="cacheVars">The CacheVars schema object to evaluate (typically from TreeNode.CacheVars).</param>
+        public async Task SetCache(dynamic cacheVars)
```
TravisOnGit, line 290: Let's add a UT that does the schema validation pattern:
await treeWalkerSession.SetCache(forgeTreeNode.CacheVars);
await treewalkerSession.EvaluateDynamicProperty(childSelector, typeof(bool));

Look around if any other UTs are manually calling EvaluateDynamicProperty already and add the UT near there.
Add a comment on the UT describing this is a way to add a safety check to confirm a TreeNode is properly using CacheVars. Confirm CacheVars and ShouldSelect statements evaluate without issue, confirming they are using the named properties correctly.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -571,25 +571,17 @@ public static class ForgeSchemaHelper
...
-                            ""Root_CollectDiagnosticsAction"": {
-                                ""Action"": ""CollectDiagnosticsAction"",
-                                ""Input"": {
-                                    ""Command"": ""TheCommand""
-                                }
-                            }
-                        },
-                        ""CacheVariables"": {
+                        ""Type"": ""Selection"",
+                        ""CacheVars"": {
```
TravisOnGit, line 584: Add a failing schema that tries to set "CacheVars": "UnexpectedStringNotADictionary". This should fail ForgeSchemaValidationRules since it's not a dictionary, even though it's a technically valid ForgeTree.

You can add the schema to jsonSchemaFailureBlacklist in ForgeSchemaValidationTests.

TravisOnGit (commented):
Left comments on iteration 9.

Forge.TreeWalker.UnitTests/test/ForgeSchemaHelper.cs
```diff
@@ -691,11 +691,11 @@ public static class ForgeSchemaHelper
                     ""SecondNode"": {
                         ""Type"": ""Selection"",
                         ""CacheVars"": {
-                            ""secondNodeCheck"": ""C#|Cache == null ? \""isolated\"" : \""leaked\""""
+                            ""secondNodeVar"": ""C#|\""present\""""
                         },
                         ""ChildSelector"": [
                             {
-                                ""ShouldSelect"": ""C#|Cache.secondNodeCheck == \""isolated\"""",
+                                ""ShouldSelect"": ""C#|Cache.secondNodeVar == \""present\"" && Cache.firstNodeVar == null"",
```
TravisOnGit, line 698: Does calling 'Cache.firstNodeVar == null' not throw an exception?

TravisOnGit (approved):
Approved iteration 10!

